PR #886, opened 2024-02-23
Update name displayed in account drawer on rename

asiia-trilitech:
## Proposed changes

[Task link](https://app.asana.com/0/1205770721172203/1206597776005791/f)

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|        |     |

## Checklist

### DIFF
--- a/src/components/AccountDrawer/index.tsx
+++ b/src/components/AccountDrawer/index.tsx
@@ -1,7 +1,7 @@
 import { useContext } from "react";
 
 import { AccountDrawerDisplay } from "./AccountDrawerDisplay";
-import { Account } from "../../types/Account";
+import { RawPkh } from "../../types/Address";
 import {
   useGetAccountAllTokens,
   useGetAccountBalance,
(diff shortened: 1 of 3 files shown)


src/components/AccountDrawer/index.tsx
```diff
@@ -24,6 +27,8 @@ export const AccountCard: React.FC<{ account: Account }> = ({ account }) => {
 
   const { openWith } = useContext(DynamicModalContext);
 
+  account = getImplicitAccountSafe(account.address.pkh)!;
+
```
serjonya-trili, line 31: If you are using the `!` operator maybe it makes sense to use the get account unsafe instead?

Also, looks like it does not make sense anymore to pass in the account, but it's better to use the PKH instead because you are going to fetch it anyway

And with that in mind it might be still useful to use the selected account context

serjonya-trili: I checked other places, yeah, it's fine to leave it as is

serjonya-trili (approved):

PR #937, opened 2024-03-06
Simplify dynamicModalContextMock usage

serjonya-trili:
this way we don't need to manually wrap our components with the context. the spying functions are always there and we don't lose the original implementation, it still renders the modal on the page


asiia-trilitech (approved):
Some ✨ typescript magic ✨ 
Thanks, Sergey!

PR #940, opened 2024-03-07
Open Beacon modal only after it is ready

serjonya-trili:
## Proposed changes

[Task link](https://app.asana.com/0/1205770721172203/1206695783870859/f)

- Beacon now supports batch operations
- Beacon modals have a dApp tile and the network so that it's easy to understand where the request comes from
- We don't open an empty modal anymore until after we have estimated the operations which we want to sign. That way we don't block the UI with a grey overlay whist it's being estimated

BeaconSignPage tests will be added in a separate PR

## Types of changes

- [x] Bugfix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] UI fix

## Steps to reproduce
you can use https://ghostnet.objkt.com/
- buy a token
- [list](https://docs.objkt.com/product/product-guides/token-buy-sell/listings) it
- re-list it with another cost 
Screenshot 2024-03-07 at 13 17 47

## Screenshots

Screenshot 2024-03-07 at 13 09 56
Screenshot 2024-03-07 at 13 14 01
Screenshot 2024-03-07 at 13 14 10

## Checklist

- [x] Documentation has been added (if appropriate)


src/utils/beacon/useHandleBeaconMessage.test.tsx
```diff
@@ -0,0 +1,378 @@
...
+        destination: mockImplicitAddress(2).pkh,
+      },
+    ];
+
+    const operations = toAccountOperations(operationDetails, account);
+
+    expect(operations).toEqual(
+      makeAccountOperations(account, account, [
+        mockTezOperation(1),
+        { ...mockTezOperation(2), recipient: mockImplicitAddress(2) },
```
asiia-trilitech, line 297: Can we update `mockTezOperation` or create another mock instead of using it like this? 
It's not very clear what's going on

src/utils/hooks/useAsyncActionHandler.ts
```diff
@@ -44,7 +44,11 @@ export const useAsyncActionHandler = () => {
     }
   };
 
-  // same as withLoadingUnsafe, but returns undefined instead of throwing an error
+  /**
+   * same as {@link handleAsyncActionUnsafe}, but returns undefined instead of throwing an error
```
asiia-trilitech, line 48: nit: "Same" + dots at the end of sentences

src/utils/beacon/useHandleBeaconMessage.tsx
```diff
@@ -0,0 +1,151 @@
...
+          case BeaconMessageType.PermissionRequest: {
+            modal = <PermissionRequestModal request={message} />;
+            break;
+          }
+          case BeaconMessageType.SignPayloadRequest: {
+            modal = <SignPayloadRequestModal request={message} />;
+            break;
+          }
+          case BeaconMessageType.OperationRequest: {
+            const signer = getAccount(message.sourceAddress) as ImplicitAccount;
```
asiia-trilitech, line 49: There was an error modal in the past - are we just throwing an error now when signer is not an owned account?

asiia-trilitech: If so - can we have a test for this case please?

asiia-trilitech (approved):
Thank you, Sergey - brilliantly done

Please check comments before submitting

src/utils/beacon/useHandleBeaconMessage.test.tsx
serjonya-trili: I inlined the values here because changing mockTezOperation is too much for this PR. but I agree it should be updated

PR #941, opened 2024-03-07
Update ErrorPage UI

asiia-trilitech:
## Proposed changes

[Task link](https://app.asana.com/0/1205770721172203/1205771213630733/f)

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|    Screenshot 2024-03-06 at 08 50 38    |  Screenshot 2024-03-13 at 14 56 27 |

https://github.com/trilitech/umami-v2/assets/150050388/1ed93c10-c5ee-45a4-bb79-2ef483592a6a

## Checklist


asiia-trilitech:
@serjonya-trili "Off-board Wallet" button off-boards user straight away, without showing the usual modal first - should I fix that & show the modal on the button click?

serjonya-trili:
> @serjonya-trili "Off-board Wallet" button off-boards user straight away, without showing the usual modal first - should I fix that & show the modal on the button click?

that's definitely a good idea 👍

src/assets/icons/Reload.tsx
```diff
@@ -0,0 +1,22 @@
...
+    width="24px"
+    height="24px"
+    fill="none"
+    viewBox="0 0 24 24"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+  >
+    <path
+      d="M16.1791 5.82018C15.5007 5.13713 14.6938 4.59511 13.8049 4.22539C12.916 3.85568 11.9627 3.66556 10.9999 3.66602C6.94828 3.66602 3.67578 6.94768 3.67578 10.9993C3.67578 15.051 6.94828 18.3327 10.9999 18.3327C14.4191 18.3327 17.2699 15.9952 18.0858 12.8327H16.1791C15.8015 13.9047 15.1005 14.8332 14.1728 15.4899C13.2452 16.1467 12.1366 16.4994 10.9999 16.4993C7.96578 16.4993 5.49995 14.0335 5.49995 10.9993C5.49995 7.96518 7.96578 5.49935 10.9999 5.49935C12.5216 5.49935 13.8783 6.13185 14.8683 7.13102L11.9166 10.0827H18.3333V3.66602L16.1791 5.82018Z"
+      fill={colors.blue}
```
serjonya-trili, line 16: please move it to the level above so that we could override it with props if needs be

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
-    link.href = jsonString;
-    link.download = "data.json";
+            <Button width="100%" onClick={onOpenOffboardingModal} size="lg" variant="warning">
+              Off-board Wallet
+            </Button>
 
-    link.click();
-  };
+            <Button width="100%" size="lg" variant="tertiary">
+              <Link
```
serjonya-trili, line 49: please add `_hover={{textDecoration: "none"}}` so that it's not underlined on hover

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
-          </Button>
-          <Button marginRight={2} onClick={onOffboard} variant="warning">
-            Offboard
-          </Button>
-          <Button variant="tertiary">
-            <Link
-              href={`mailto:umami-support@trili.tech?subject=Umami V2 feedback&body=${feedbackEmailBodyTemplate}`}
-              rel="noopener noreferrer"
-              target="_blank"
+            <Button
```
serjonya-trili, line 58: I am not sure if this one is a button. Please confirm with Bahador. As I understand it should be something like [that](https://github.com/trilitech/umami-v2/blob/main/src/components/Onboarding/derivationPath/DerivationPath.tsx#L79-L102)

serjonya-trili: please also check the spacing. it is too far away from the button above

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
+        padding="40px"
+        background={colors.gray[900]}
+        border="1px"
+        borderColor={colors.gray[700]}
+        borderRadius="8px"
+        boxShadow="0px 0px 30px rgba(0, 0, 0, 0.30)"
+      >
+        <ModalContentWrapper
+          icon={<NoticeIcon />}
+          subtitle="Please refresh the app or use one of the following options:"
```
serjonya-trili, line 36: the titles have smaller text sizes than they have to and the header block is narrower on the design, please also verify that the spacing between it and the first button is correct

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
   };
 
-  const onBackup = () => {
-    const jsonString = `data:text/json;chatset=utf-8,${encodeURIComponent(
-      JSON.stringify(window.localStorage)
-    )}`;
+  return (
+    <Center height="100vh" padding="60px" backgroundImage={BackgroundImage} backgroundSize="cover">
+      <Box
+        width="100%"
```
serjonya-trili, line 23: not sure about this one. looks like it's simpler to set `width="460px"`
also, looks like it should be `480px` according to the design

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
+        maxWidth="460px"
+        height="100%"
```
serjonya-trili, line 25: this is kinda the default behaviour IIRC

src/components/ErrorPage.tsx
```diff
@@ -1,59 +1,75 @@
...
+        maxHeight="480px"
```
serjonya-trili, line 26: I don't think we should have a maxHeight at all because what if we exceed it? it should grow naturally with the amount of content it holds (unless we want to have a scrollbar)

src/components/ErrorPage.tsx
asiia-trilitech: Messaged Bahador regarding this

src/components/ErrorPage.tsx
asiia-trilitech: In my opinion we should use the same size of header & subtitle for all similar modals. However, Figma contains different font sizes for different modals atm.

Issue with the header block width is also present with the other modals (e.g. for onboarding)

I've sent a question to Bahador regarding this as well

src/components/ErrorPage.tsx
asiia-trilitech: According to Bahador:
> The Refresh should be a button. Please follow the same spacing above and below the Refresh button as the design.

src/components/ErrorPage.tsx
asiia-trilitech: According to Bahador:

> Please follow the same logic as the other modals (e.g. onboarding). So use the same font size, and the same width as other modals.

The description width should be constant & smaller than modal context's width

src/components/ErrorPage.tsx
asiia-trilitech: Actually, just discussed this offline - it should be a link, the color should be changed on hover same way as for "Save Backup" button

serjonya-trili: it's still a bit further away than the other buttons. I don't think you need this top margin
Screenshot 2024-03-13 at 10 22 14

asiia-trilitech: Let me check this with Bahador one more time just in case

asiia-trilitech:
@serjonya-trili updated the code to match the Figma & Bahador's response. Please take a look when you have time 🙂 

According to Bahador, the margin between "Report Error" and "Refresh" should be bigger (same as in Figma)

serjonya-trili (approved):

PR #956, opened 2024-03-12
Fix duplicating key error for peers

asiia-trilitech:
## Proposed changes

DApp might have the same name, but different beacon instances.
E. g. on mainnet & ghostnet, or when trying to connect from different browsers.

SenderId is unique for each beacon instance of a dApp, so using it as a key solves this problem.

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Steps to reproduce

Connect to the same dApp from different browsers. Check console for errors.

## Screenshots

| Before | Now |
| ------ | --- |
|    Screenshot 2024-03-12 at 14 00 38    |   Screenshot 2024-03-12 at 14 02 27  |

## Checklist

### DIFF
--- a/src/utils/beacon/BeaconPeers.tsx
+++ b/src/utils/beacon/BeaconPeers.tsx
@@ -72,7 +72,7 @@ const PeersDisplay = ({
 }) => (
   <Box>
     {peerInfos.map(peerInfo => (
-      <Fragment key={peerInfo.name}>
+      <Fragment key={peerInfo.senderId}>
         <Divider />
         <PeerRow onRemove={() => removePeer(peerInfo)} peerInfo={peerInfo} />
       </Fragment>


src/utils/beacon/BeaconPeers.test.tsx
```diff
@@ -88,6 +88,34 @@ describe("<BeaconPeers />", () => {
       }
     });
 
+    it("resolves duplicating name for different peer connections", async () => {
```
serjonya-trili, line 91: Imo the test is redundant. Have you seen it fail?

asiia-trilitech: Yes, with the previous key `key={peerInfo.name}``

I can remove the test if you think we don't need it

serjonya-trili (approved):

PR #983, opened 2024-03-18
Upgrade redux, react-redux, and redux-toolkit

serjonya-trili:


asiia-trilitech (approved):
Tested manually - all seems to be working correctly

PR #1000, opened 2024-03-19
Refactor code related to labels & contacts uniqueness

asiia-trilitech:
## Proposed changes

Prep for the following task: [Task link](https://app.asana.com/0/1205770721172203/1205770721303329/f)

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix

## Checklist

- [x] Documentation has been added (if appropriate)
- [x] All TODOs have a corresponding task created (and the link is attached to it)


src/components/ContactModal.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { contact1, contact2 } from "../mocks/contacts";
+import { mockImplicitAddress } from "../mocks/factories";
+import { fireEvent, render, screen, waitFor } from "../mocks/testUtils";
+import { contactsActions } from "../utils/redux/slices/contactsSlice";
+import { store } from "../utils/redux/store";
+
+describe("UpsertContactModal", () => {
+  describe("on adding contact", () => {
+    describe.each([
+      { testCase: "new contact", renderComponent: () => render(<UpsertContactModal />) },
```
serjonya-trili, line 11: you don't have to use `render` here, you can just leave the element itself and use render in the test case

src/components/ContactModal.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+      it("adds contact to address book", async () => {
+        store.dispatch(contactsActions.upsert(contact2));
+        renderComponent();
+
+        // Set name
+        fireEvent.change(screen.getByLabelText("Name"), {
+          target: { value: "Test Contact" },
+        });
+        fireEvent.blur(screen.getByLabelText("Name"));
```
serjonya-trili, line 77: do you need the `blur` event here? if not maybe you could use `userEvent` instead for writing the text into the input?

src/components/ContactModal.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+      // Update name
+      fireEvent.change(screen.getByLabelText("Name"), {
+        target: { value: "Updated Name" },
+      });
+      fireEvent.blur(screen.getByLabelText("Name"));
+
+      // Submit
+      await waitFor(() => expect(screen.getByTestId("confirmation-button")).toBeEnabled());
+      fireEvent.click(screen.getByTestId("confirmation-button"));
```
serjonya-trili, line 220: again, please prefer `userEvent` to `fireEvent`

src/components/ContactModal.tsx
```diff
@@ -20,27 +20,30 @@ import { CopyableAddress } from "./CopyableText";
...
-  const getAccount = useGetOwnedAccountSafe();
   const { isOpen, onClose } = useContext(DynamicModalContext);
 
-  const onSubmitContact = (newContact: Contact) => {
-    if (getAccount(newContact.pkh)) {
-      return;
-    }
-    dispatch(contactsActions.upsert(newContact));
+  // When editing existing contact, its name & pkh are known.
+  const isEdit = contact !== undefined && contact.pkh != "" && contact.name !== "";
```
serjonya-trili, line 43: so, if contact is not undefined then why we check that pkh and name are non-empty? they must be, don't they? also, please don't use `!=`, use `!==` instead

src/components/ContactModal.tsx
```diff
@@ -110,29 +88,39 @@ export const UpsertContactModal: FC<{
...
+            )}
           </FormControl>
           <FormControl isInvalid={!!errors.pkh} marginY={5}>
             <FormLabel>Address</FormLabel>
             <Input
               type="text"
               {...register("pkh", {
                 required: "Address is required",
-                validate: validatePkh,
+                validate: isEdit ? () => true : validatePkh,
```
serjonya-trili, line 101: well, it's an optimisation, but not very needed tbh. we won't gain much performance gain. it's fine to always validate it

src/utils/hooks/contactsHooks.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      { type: "ledger", pkh: mockLedgerAccount(0).address.pkh },
+      { type: "social", pkh: mockSocialAccount(1).address.pkh },
+      { type: "secret_key", pkh: mockSecretKeyAccount(2).address.pkh },
+      { type: "mnemonic", pkh: mockMnemonicAccount(3).address.pkh },
+      { type: "multisig", pkh: mockMultisigAccount(0).address.pkh },
+    ])("fails for address used for $type account", ({ pkh }) => {
+      store.dispatch(contactsActions.reset());
+      store.dispatch(accountsSlice.actions.addAccount(mockLedgerAccount(0)));
+      store.dispatch(accountsSlice.actions.addAccount(mockSocialAccount(1)));
+      store.dispatch(accountsSlice.actions.addAccount(mockSecretKeyAccount(2)));
```
serjonya-trili, line 133: there is a neat test helper called just `addAccount` which should shorten all of these

src/utils/hooks/contactsHooks.ts
```diff
@@ -1,21 +1,35 @@
...
-  };
+  return (pkh: string) => pkh in contacts;
 };
 
 export const useGetContactName = () => {
   const contacts = useContacts();
   return (pkh: string) => (pkh in contacts ? contacts[pkh].name : undefined);
 };
+
+export const useValidatePkh = () => {
```
serjonya-trili, line 22: it's better to be split into several validators. for example, if we use it in the address autocomplete as a destination for a tez transfer, why do we care if it exists in our accounts?
or, if it's intended to be used just for contacts, then let's call it `useValidateNewContactPkh` or something

src/components/ContactModal.tsx
asiia-trilitech: The check is needed here at least for the name. 

There are 3 existing use cases for this modal. 

Creating new contact:
- undefined contact (creating from scratch)
- contact with pkh and empty name (adding used address to contacts)

Editing existing contact:
- contact with both pkh and non-empty name

src/components/ContactModal.tsx
asiia-trilitech: Without this fix, the check fails for edit case - cause pkh already exists in contacts 🙂

src/components/ContactModal.tsx
serjonya-trili: what about `const isEdit = contact?.pkh && contact?.name` then?

asiia-trilitech: Maybe I got something wrong, but it produced the wrong type
`(contact?.pkh && contact.name) ? true : false` worked though 🙂

serjonya-trili: then `const isEdit = !!(contact?.pkh && contact.name)` please

src/components/ContactModal.test.tsx
asiia-trilitech: It doesn't work with just `user.type`. For some reason `user.type` doesn't trigger whatever `blur` triggers.
But I managed to make it work with clicking outside of the text input

src/utils/hooks/contactsHooks.test.ts
asiia-trilitech: Good one, fixed

src/components/ContactModal.test.tsx
serjonya-trili: if you need the blur even then please stick to the `fireEvent`

src/utils/hooks/contactsHooks.ts
asiia-trilitech: I think we only need a separate check for the contracts - renamed

src/utils/hooks/contactsHooks.ts
```diff
@@ -1,21 +1,35 @@
...
+export const useValidateNewContactPkh = () => {
+  const addressExistsInContacts = useAddressExistsInContacts();
+  const getAccount = useGetOwnedAccountSafe();
+
+  return (pkh: string) => {
+    if (!isAddressValid(pkh)) {
+      return "Invalid address";
+    }
+    if (getAccount(pkh)) {
+      return "Address already used in accounts";
```
serjonya-trili, line 31: Address *is* ...
same below please

serjonya-trili (approved):

src/utils/hooks/contactsHooks.ts
asiia-trilitech: Thank you, fixed

PR #1001, opened 2024-03-20
Memoize handleAsyncAction functions

serjonya-trili:
## Proposed changes

- This improves performance because the functions do not get recreated all the time
- It allows to use those functions in useEffect's or similar hooks' dependency arrays
- It doesn't depend on the render cycle anymore and should be much safer in terms of race conditions
- fixed a bug in multisig creation part where the same handleAsyncAction was reused which might not let the contract name get assigned

### DIFF
--- a/src/utils/hooks/useAsyncActionHandler.test.ts
+++ b/src/utils/hooks/useAsyncActionHandler.test.ts
@@ -1,3 +1,5 @@
+import { waitFor } from "@testing-library/react";
+
 import { useAsyncActionHandler } from "./useAsyncActionHandler";
 import { act, renderHook } from "../../mocks/testUtils";
 import { mockToast } from "../../mocks/toast";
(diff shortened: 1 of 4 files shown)


src/utils/hooks/useAsyncActionHandler.test.ts
asiia-trilitech, line 1: Looks like this should now be fixed https://app.asana.com/0/1205770721172203/1206776085221902/f

src/utils/hooks/useAsyncActionHandler.ts
```diff
@@ -1,58 +1,67 @@
...
-import { useState } from "react";
+import { useCallback, useRef, useState } from "react";
 
 import { getErrorContext } from "../getErrorContext";
 import { useAppDispatch } from "../redux/hooks";
 import { errorsSlice } from "../redux/slices/errorsSlice";
 
+/**
+ * Hook for gracefully handling async actions.
+ * If an action fails, it'll show a toast with a corresponding message
```
asiia-trilitech, line 10: nit: please add empty line after the first sentence.

src/utils/hooks/useAsyncActionHandler.ts
```diff
@@ -1,58 +1,67 @@
...
-   * Same as {@link handleAsyncActionUnsafe}, but returns undefined instead of throwing an error.
-   *
-   * NOTE: might be helpful in tests to remove the catch part of this function so that you'd see all the exceptions.
-   *  */
-  const handleAsyncAction = async <T>(
-    fn: () => Promise<T>,
-    toastOptions?: UseToastOptions | ((error: any) => UseToastOptions)
-  ): Promise<T | void> => handleAsyncActionUnsafe(fn, toastOptions).catch(() => {});
+  // NOTE: might be helpful in tests to remove the catch part of this function so that you'd see all the exceptions.
+  const handleAsyncAction = useCallback(
```
asiia-trilitech, line 58: There is useCallback for `handleAsyncActionUnsafe`, do we need another one here?

src/utils/hooks/useAsyncActionHandler.ts
```diff
@@ -1,58 +1,67 @@
...
-  /**
-   * Same as {@link handleAsyncActionUnsafe}, but returns undefined instead of throwing an error.
-   *
-   * NOTE: might be helpful in tests to remove the catch part of this function so that you'd see all the exceptions.
-   *  */
-  const handleAsyncAction = async <T>(
-    fn: () => Promise<T>,
-    toastOptions?: UseToastOptions | ((error: any) => UseToastOptions)
-  ): Promise<T | void> => handleAsyncActionUnsafe(fn, toastOptions).catch(() => {});
+  // NOTE: might be helpful in tests to remove the catch part of this function so that you'd see all the exceptions.
```
asiia-trilitech, line 57: nit: maybe add something like "for debugging purposes"?

asiia-trilitech (commented):
Thanks for the fix, Sergey!

LGTM 👍

src/utils/hooks/useAsyncActionHandler.ts
serjonya-trili: yes, because otherwise we'd keep creating new functions on each render & we wouldn't be able to use it as one of useEffect's dependencies

asiia-trilitech (approved):
👍

PR #1010, opened 2024-03-21
Fix autocomplete suggestions scrolling

serjonya-trili:
## Proposed changes

It wasn't working when a drawer is open, but was when you had just a modal window open. Now both cases work. Most likely was broken due to a charka dependency update


asiia-trilitech (approved):
Checked manually - all good

Thank you for the quick fix 🔥 🔥 🔥

PR #1031, opened 2024-03-26
Optimize useAccountsFilter hook

serjonya-trili:
## Proposed changes

Now the selected accounts object is cached and won't change until we actually update the list of selected accounts. Since useGetOperations relies on it, it's helpful to use the object as a dependency in useEffect or similar hooks and ovoid unnecessary requests

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix

## Checklist


asiia-trilitech (approved):
Thanks for the optimization, Sergey! 👍 👍 👍 

Code looks good. And I checked the change manually as well.

As for the test update - old version of the filter sorted address pills by name (which seems like a wrong behaviour to me).
Current version does not do the sorting, hence the required changes.

PR #1057, opened 2024-04-03
Fix unrelated operations rendering

serjonya-trili:
## Proposed changes

Some contract calls may trigger multiple operations. Some of those may include operations which do not involve the caller at all, so those are neither incoming nor outgoing

Also, sometimes we may burn tokens/tez. So we don't really have the target. now we correctly handle such cases

## Types of changes

- [x] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|        |     |

## Checklist


src/components/OperationTile/useGetOperationDestination.ts
```diff
@@ -0,0 +1,31 @@
...
+  sender: RawPkh | null | undefined,
+  receiver: RawPkh | null | undefined
+) => {
+  const isSenderOwned = useIsOwnedAddress(sender || "");
+  const isReceiverOwned = useIsOwnedAddress(receiver || "");
+
+  if (!isSenderOwned && !isReceiverOwned) {
+    return "unrelated";
+  }
+
```
asiia-trilitech, line 25: The logic seems to be simplified quite a lot?

- Maybe we should also handle separately the case when both sender and receiver are owned?

- For the account drawer the "current" account should take precedent for choosing the sign?

src/utils/tezos/fetch.ts
```diff
@@ -60,7 +60,7 @@ export type TransactionOperation = tzktApi.TransactionOperation & {
   counter: number;
   type: "transaction";
   sender: TzktAlias;
-  target: TzktAlias;
+  target: tzktApi.Alias;
```
asiia-trilitech, line 63: Sorry, not sure I understand this bit

src/components/OperationTile/TransactionTile.tsx
```diff
@@ -15,21 +15,30 @@ import { TransactionOperation } from "../../utils/tezos";
...
+  );
   const amount = prettyTezAmount(String(operation.amount));
   const showToAddress = useShowAddress(operation.target.address);
   const showFromAddress = useShowAddress(operation.sender.address);
   // if you send assets between your own accounts you need to see at least one address
   const showAnyAddress = !showToAddress && !showFromAddress;
 
-  const titleColor = isOutgoing ? colors.orange : colors.green;
-  const sign = isOutgoing ? "-" : "+";
+  const titleColor =
```
asiia-trilitech, line 28: nit: code for determining color & sign is duplicated. Maybe we can move it to operation destination?

src/components/OperationTile/TransactionTile.tsx
```diff
@@ -15,21 +15,30 @@ import { TransactionOperation } from "../../utils/tezos";
 import { AddressPill } from "../AddressPill/AddressPill";
 
 export const TransactionTile: React.FC<{ operation: TransactionOperation }> = ({ operation }) => {
-  const isOutgoing = useIsOutgoingOperation(operation.sender.address);
+  const operationDestination = useGetOperationDestination(
+    operation.sender.address,
```
asiia-trilitech, line 19: Do we need "?" here and/or in the next line?

src/components/OperationTile/useGetOperationDestination.ts
serjonya-trili: when they both owned it's still an outgoing operation because you lose the fee
"current" account logic is too complicated and while it's nice in a way, it's also a bit confusing to see the same operation as incoming & outgoing in 2 different places

PR #1058, opened 2024-04-04
Change loading animations

serjonya-trili:
https://github.com/trilitech/umami-v2/assets/129749432/7362698e-05b2-43cd-81f0-04c95b8d124b


src/views/operations/OperationsView.tsx
```diff
@@ -31,17 +31,20 @@ export const OperationsView = () => {
...
     <Flex flexDirection="column" height="100%">
       <TopBar title="Operations" />
       {accountsFilter}
-      {isLoading && isFirstLoad && loadingElement}
+
+      <Center display={isLoading && isFirstLoad ? "flex" : "none"} height="100%">
+        <Image
+          width="150px"
+          height="75px"
+          marginBottom="136px"
```
asiia-trilitech, line 43: Why do we need marginBottom here? 🤔

asiia-trilitech (approved):
Thank you, Sergey - looks good!

Please check comments before submitting 🙂

src/views/operations/OperationsView.tsx
serjonya-trili: otherwise, it's not centerred against the whole screen, but rather only the part under the topbar

PR #1060, opened 2024-04-04
Safer offboarding

serjonya-trili:
## Proposed changes

Sometimes beacon would throw an error if we only cleaned the local storage.
also, it might be that redux persist flushes the state to the local storage right after we've just cleaned it.
also, `useReset` has never been a hook

## Types of changes

- [x] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix


asiia-trilitech (approved):
Checked manually - all works
Thanks for the fix, Sergey!

PR #1077, opened 2024-04-08
Remove tez transfers from the assets slice

serjonya-trili:


asiia-trilitech:
Why are we removing this data from the slice?

serjonya-trili:
> Why are we removing this data from the slice?

Because it's not used anywhere

PR #1194, opened 2024-04-30
Migrate to tezbox image

serjonya-trili:
## Proposed changes
It is better maintained than flextesa


asiia-trilitech (approved):
Thank you, Sergey - the change looks good 👍

PR #1258, opened 2024-05-16
Parse beacon payload

OKendigelyan:
## Proposed changes

[Task link](https://app.asana.com/0/1205770721172203/1206727614749706/f)

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|image|image|

## Checklist

### DIFF
--- a/src/utils/beacon/SignPayloadRequestModal.tsx
+++ b/src/utils/beacon/SignPayloadRequestModal.tsx
@@ -17,6 +17,7 @@ import {
 import { TezosToolkit } from "@taquito/taquito";
 import React, { useContext } from "react";
 
+import { decodePayload } from "./decodePayload";
 import { WalletClient } from "./WalletClient";
 import { DynamicModalContext } from "../../components/DynamicModal";
 import { SignButton } from "../../components/SendFlow/SignButton";
@@ -43,7 +44,10 @@ export const SignPayloadRequestModal: React.FC<{
 
     await WalletClient.respond(response);
 
-    toast({ description: "Successfully submitted Beacon operation", status: "success" });
+    toast({
+      description: "Successfully submitted Beacon operation",
+      status: "success",
+    });
     onClose();
   };
 
@@ -59,6 +63,8 @@ export const SignPayloadRequestModal: React.FC<{
           {`${request.appMetadata.name}/dApp Pairing Request`}
         </Heading>
         <Box
+          overflowY="auto"
+          maxHeight="300px"
           padding="15px"
           border="1px solid"
           borderColor={colors.gray[500]}
@@ -66,7 +72,7 @@ export const SignPayloadRequestModal: React.FC<{
           backgroundColor={colors.gray[800]}
         >
           <Text color={colors.gray[450]} size="md">
-            {request.payload}
+            {decodePayload(request.payload)}
           </Text>
         </Box>
       </ModalBody>
--- a/src/utils/beacon/decodePayload.test.ts
+++ b/src/utils/beacon/decodePayload.test.ts
@@ -1,0 +1,37 @@
+import { decodePayload } from "./decodePayload";
+
+describe("decodePayload", () => {
+  it("decodes a valid payload with '0501' prefix and padding", () => {
+    const payload =
+      "05010000004254657a6f73205369676e6564204d6573736167653a206d79646170702e636f6d20323032312d30312d31345431353a31363a30345a2048656c6c6f20776f726c6421";
+    const expected = "Tezos Signed Message: mydapp.com 2021-01-14T15:16:04Z Hello world!";
+    expect(decodePayload(payload)).toEqual(expected);
+  });
+
+  it("decodes a payload without '0501' prefix", () => {
+    const payload =
+      "54657a6f73205369676e6564204d6573736167653a206d79646170702e636f6d20323032312d30312d31345431353a31363a30345a2048656c6c6f20776f726c6421";
+    const expected = "Tezos Signed Message: mydapp.com 2021-01-14T15:16:04Z Hello world!";
+    expect(decodePayload(payload)).toEqual(expected);
+  });
+
+  it("returns original payload if length is invalid", () => {
+    const payload = "05010000005254657a6f73205369676e6564204d6573736167653a20696e76616c69642e";
+    expect(decodePayload(payload)).toEqual(payload);
+  });
+
+  it("returns original payload if an error occurs during decoding", () => {
+    const invalidPayload = "0501000000";
+    expect(decodePayload(invalidPayload)).toEqual(invalidPayload);
+  });
+
+  it("handles an empty payload", () => {
+    const emptyPayload = "";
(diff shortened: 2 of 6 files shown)


src/utils/beacon/payloadDecoder.ts
```diff
@@ -0,0 +1,18 @@
+const hexToBytes = (hex: string): Uint8Array => {
```
serjonya-trili, line 1: I bet you can use `hex2buf` from taquito

src/utils/beacon/payloadDecoder.ts
```diff
@@ -0,0 +1,18 @@
...
+  const bytes = new Uint8Array((hex.length - 12) / 2);
+
+  for (let i = 12; i < hex.length; i += 2) {
+    bytes[(i - 12) / 2] = parseInt(hex.slice(i, i + 2), 16);
+  }
+
+  return bytes;
+};
+
+export const payloadDecoder = (payload: string): string => {
```
serjonya-trili, line 11: this function can be shortened to just
```typescript
export const payloadDecoder = (payload: string) =>
  new TextDecoder("utf-8").decode(hex2buf(payload));
```

src/utils/beacon/SignPayloadRequestModal.tsx
```diff
@@ -59,14 +63,16 @@ export const SignPayloadRequestModal: React.FC<{
...
+          maxHeight="300px"
           padding="15px"
           border="1px solid"
           borderColor={colors.gray[500]}
           borderRadius="4px"
           backgroundColor={colors.gray[800]}
         >
           <Text color={colors.gray[450]} size="md">
-            {request.payload}
+            {payloadDecoder(request.payload)}
```
serjonya-trili, line 75: it might be that we cannot parse the payload (it's just an arbitrary text converted to a hex string, you can pass something like 0501000000000000, which cannot be converted to a string)
so, in such case, the function should throw an exception, and here we should display the actual payload

serjonya-trili (changes requested):

src/utils/beacon/payloadDecoder.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      index += 1;
+    }
+  }
+
+  try {
+    const string = new TextDecoder("utf-8").decode(hex2buf(payload.slice(index)));
+
+    return string.length > 0 ? string : payload;
```
serjonya-trili, line 41: `return string || payload` maybe?

src/utils/beacon/payloadDecoder.ts
```diff
@@ -0,0 +1,45 @@
+import { hex2buf } from "@taquito/utils";
+
+const PAYLOAD_PADDING = "0501";
```
serjonya-trili, line 3: please add a link where you took it from (either taquito or beacon would suffice)

src/utils/beacon/decodePayload.ts
```diff
@@ -0,0 +1,54 @@
...
+      index += 1;
+    }
+  }
+
+  try {
+    const string = new TextDecoder("utf-8").decode(
+      hex2buf(payload.slice(index))
+    );
+
+    return string.length > 0 ? string : payload;
```
serjonya-trili: not sure if you've seen. I think `return string || payload` looks a bit better

src/utils/beacon/decodePayload.test.ts
```diff
@@ -0,0 +1,37 @@
+import { decodePayload } from "./decodePayload";
+
+describe("decodePayload", () => {
+  it("should decode a valid payload with '0501' prefix and padding", () => {
```
serjonya-trili, line 4: `it("should decode ..."` is always longer than `it("decodes ...` so it's better to avoid the word should unless really needed

src/utils/beacon/decodePayload.ts
```diff
@@ -0,0 +1,52 @@
...
+const PAYLOAD_PADDING = "0501";
+
+/**
+ * Decodes a sign request payload string.
+ *
+ * @param payload - The payload to decode.
+ * @returns The decoded string, or the original payload if decoding fails.
+ * @see {@link https://taquito.io/docs/signing/} for more info.
+ */
+export const decodePayload = (payload: string): string => {
```
serjonya-trili, line 12: you could also change the type of payload to `SignPayloadRequestOutput['payload']`

that way, you can clearly see and, what's more important, navigate to the actual type definition

serjonya-trili (approved):
please consider the remaining comments
other than that, great job 👍

asiia-trilitech:
Hey Oleg!

A small note: the `Task link` in the description should go to Asana task (if applicable) 🙂 
I assume for this PR it's this one? https://app.asana.com/0/1205770721172203/1206727614749706/f

src/utils/beacon/decodePayload.ts
```diff
@@ -0,0 +1,52 @@
+import { hex2buf } from "@taquito/utils";
+
+const PAYLOAD_PADDING = "0501";
```
asiia-trilitech, line 3: Should it be `PAYLOAD_PREFIX` or `OPTIONAL_PAYLOAD_PREFIX` considering there might be zeros as padding as well?

asiia-trilitech: A comment in code for what this prefix means/does would also be good 🙂

src/utils/beacon/decodePayload.ts
```diff
@@ -0,0 +1,52 @@
...
+    while (index < payload.length && payload[index] === "0") {
+      index += 1;
+    }
+
+    let rawBytesLength = "";
+    let bytesLength = 0;
+    const paddingLength = index;
+
+    while (index < payload.length && payload[index] !== "0") {
+      rawBytesLength += payload[index];
```
asiia-trilitech, line 27: `rawBytesLength.length` seems a bit confusing to me

Please consider renaming vars to maybe something like

`rawBytesLength` => `payloadByteLengthRaw`
`bytesLength` => `payloadByteLength`

src/utils/beacon/decodePayload.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (bytesLength * 2 > remainingPayloadLength) {
+        return payload;
+      }
+
+      index += 1;
+    }
+  }
+
+  try {
+    const string = new TextDecoder("utf-8").decode(hex2buf(payload.slice(index)));
```
asiia-trilitech, line 46: [nit]

IIUC lines 13-43 are basically searching for the payload start index.

To improve readability maybe we can move them to a separate private method called something like `getPayloadHexBytes`, which
- returns payload.slice(index) to go into `TextDecoder("utf-8").decode(hex2buf())`
- throw error  or return null / empty string in case of incorrect length

asiia-trilitech (approved):
Thanks Oleg - the code looks good to me 🔥 🔥 🔥 

My comments are optional (as they do not affect the execution), but please consider applying them

PR #1341, opened 2024-06-06
Fix contract origination with Michelson storage

serjonya-trili:
## Proposed changes

here's [the reason why](https://taquito.io/docs/originate#b-initializing-storage-using-a-plain-michelson-expression-for-initial-storage)

## Types of changes

- [x] Bugfix

## Steps to reproduce
try to deploy a token [here](https://tezos-homebase.io/creator/deployment/summary)


OKendigelyan:
@serjonya-trili LGTM including one small comment 👍

PR #1372, opened 2024-06-17
Create umami web app template

OKendigelyan:
## Proposed changes

[Task link]()

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|        |     |

## Checklist


apps/desktop/jest.config.mjs
```diff
@@ -0,0 +1,7 @@
+import config from "@umami/jest-config";
+
```
serjonya-trili, line 2: Wondering why not ts?

apps/web/tsconfig.e2e.json
```diff
@@ -0,0 +1,15 @@
+{
```
serjonya-trili, line 1: I was thinking of moving e2e tests to its own package to avoid this. WDYT?

apps/web/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": { "types": ["vite/client"] },
+  "extends": "@umami/typescript-config/tsconfig.json",
+  "include": ["src", "vite.config.ts", "jest.config.js"],
```
serjonya-trili, line 4: I think we should stick to jest.config.ts instead, otherwise, as I understand, this inclusion is not helpful at all

packages/typescript-config/tsconfig.json
```diff
@@ -15,6 +15,9 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react-jsx",
-    "incremental": true
+    "incremental": true,
+    "paths": {
+      "@umami/jest-config": ["../../packages/jest-config/*"]
```
serjonya-trili, line 20: Hmm, why?

OKendigelyan: The paths option allows typescript to resolve imported internal packages

apps/web/tsconfig.e2e.json
OKendigelyan: nice idea to clean up the mess with cjs/mjs

apps/desktop/jest.config.mjs
serjonya-trili: It used to be a typescript file. And all the perks you get with typescript compiler included

apps/desktop/.eslintignore
```diff
@@ -0,0 +1,2 @@
+.eslintrc.*
```
serjonya-trili, line 1: Hmm, why?

OKendigelyan: It's a strait-forward solution to prevent ESlint from such kind of issues. Could be solved in a more elegant way in the future
image

apps/web/package.json
```diff
@@ -11,31 +11,63 @@
     "test:watch": "cross-env DEV=true yarn test --watch"
   },
   "dependencies": {
+    "@chakra-ui/anatomy": "2.2.2",
+    "@chakra-ui/cli": "^2.4.1",
```
serjonya-trili, line 15: that's definitely a dev dependency; please also adjust others too
in the desktop app, all the deps are dev deps because of how electron-builder works

PR #1401, opened 2024-06-21
Move auth into a separate package

serjonya-trili:
## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix


packages/typescript-config/tsconfig.json
```diff
@@ -10,11 +10,11 @@
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
     "module": "ESNext",
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
```
asiia-trilitech, line 13: What is the reasoning behind modeResolution change?

asiia-trilitech: And for the removing `incremental: true` 🙂

asiia-trilitech (approved):
LGTM - thank you for the change, Sergey!

.yarn/releases/yarn-4.3.1.cjs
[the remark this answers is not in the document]
serjonya-trili: yes

packages/typescript-config/tsconfig.json
serjonya-trili: TL; DR; it doesn't work otherwise

PR #1430, opened 2024-06-28
Add account block

OKendigelyan:
## Proposed changes

[Task link](https://app.asana.com/0/1207041888417083/1207635059795807/f)

## Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

image

## Checklist


apps/web/src/components/AccountCard/AccountBalance.tsx
```diff
@@ -0,0 +1,25 @@
+import { AddIcon, ArrowDownIcon } from "@chakra-ui/icons";
+import { Box, Button, Flex, IconButton, Text } from "@chakra-ui/react";
+
+export const AccountBalance = () => (
+  <Box>
+    <Box marginBottom="4">
```
serjonya-trili, line 6: let's use actual pixels please and in other places too

apps/web/src/components/AccountCard/AccountBalance.tsx
```diff
@@ -0,0 +1,25 @@
...
+        2882.675746 ꜩ
+      </Text>
+      <Text color="gray.500" fontSize="md">
+        $3603.34
+      </Text>
+    </Box>
+    <Flex alignItems="center" justifyContent="space-between">
+      <IconButton borderRadius="full" aria-label="Add" icon={<AddIcon />} size="lg" />
+      <IconButton borderRadius="full" aria-label="Send" icon={<ArrowDownIcon />} size="lg" />
+      <Button borderRadius="full" colorScheme="blue" size="lg">
```
serjonya-trili, line 20: I am not sure about the colorScheme here, I believe we should introduce our own `variant` WDYT?

apps/web/src/components/AccountCard/AccountBalance.tsx
```diff
@@ -0,0 +1,25 @@
+import { AddIcon, ArrowDownIcon } from "@chakra-ui/icons";
+import { Box, Button, Flex, IconButton, Text } from "@chakra-ui/react";
+
+export const AccountBalance = () => (
+  <Box>
+    <Box marginBottom="4">
+      <Text marginBottom="1" color="gray.500" fontSize="sm">
```
serjonya-trili, line 7: the problem with those string color literals is that they aren't type-checked. it's too easy to make a mistake and not notice it. like gray vs grey, gray.500 vs gray.5000.
also, in other places we use `mode(light.grey[50], dark.grey[50])(colorMode)` which IIRC does exactly the same

so, I'd suggest to stick to the 2nd option OR we could introduce our own hook like:

```typescript
import { useColorMode } from "@chakra-ui/react";
import { mode } from "@chakra-ui/theme-tools";
import { useCallback } from "react";

const useColor = () => {
  const colorMode = useColorMode();

  return useCallback(
    (light: number, dark?: number) =>
      mode(light.grey[light], dark.grey[dark || light])(colorMode),
    [colorMode]
  );
};
```

and then use it like this:

```typescript
const color = useColor();

<Flex background={color(100)} color={color(900, 800)}
```

apps/web/src/styles/theme.ts
```diff
@@ -45,6 +47,12 @@ const theme = extendTheme({
   components: {
     Text: {
       sizes,
+      variants: {
+        bold: {
+          fontWeight: "600",
+          color: "var(--chakra-colors-gray-900)",
```
serjonya-trili, line 53: that's not how we should define the colours. please check [this](https://v2.chakra-ui.com/docs/styled-system/semantic-tokens#conditional-tokens)

apps/web/src/vite-env.d.ts
```diff
@@ -1,2 +1,3 @@
 /// <reference types="vite/client" />
 /// <reference path="../../../global.d.ts" />
+/// <reference types="vite-plugin-svgr/client" />
```
serjonya-trili, line 3: please add an empty line and make vscode do that for your (search for new line in the settings)

apps/web/vite.config.ts
```diff
@@ -1,11 +1,45 @@
...
+        svgo: true,
+        svgoConfig: {
+          plugins: [
+            {
+              name: "preset-default",
+              params: {
+                overrides: { removeViewBox: false },
+              },
+            },
+            //     {
```
serjonya-trili, line 25: please remove :)

serjonya-trili (changes requested):

apps/web/src/assets/icons/index.ts
```diff
@@ -1,3 +1,16 @@
...
+export { default as ChevronDownIcon } from "./chevron-down.svg?react";
+export { default as FileCopyIcon } from "./file-copy.svg?react";
+export { default as PlusIcon } from "./plus.svg?react";
+export { default as ArrowDownLeftIcon } from "./arrow-down-left.svg?react";
+export { default as GoogleSocialIcon } from "./google-social.svg?react";
+export { default as FacebookSocialIcon } from "./facebook-social.svg?react";
+export { default as XSocialIcon } from "./x-social.svg?react";
+export { default as UnknownSocialIcon } from "./unknown-social.svg?react";
+export { default as LedgerSocial } from "./ledger-social.svg?react";
+export { default as MailIcon } from "./mail.svg?react";
```
serjonya-trili, line 13: are those imported? I cannot find the files

apps/web/src/styles/theme.ts
```diff
@@ -117,13 +126,59 @@ const theme = extendTheme({
...
+          color: light.blue,
+          _hover: {
+            borderColor: light.blueDark,
+            color: light.blueDark,
+          },
+          _dark: {
+            color: "#FFF",
+          },
+        },
+        iconButtonSolid: (props: StyleFunctionProps) => ({
```
serjonya-trili, line 155: looks like this would've been simpler, no? right now we are mixing an outer `_dark` with the inner ones and it's hard to follow

```typescript
iconButtonSolid: {
  bg: light.grey[100],
  color: light.grey[900],
  _hover: { bg: light.grey[200] },
  _dark: {
    bg: dark.grey[100],
    color: dark.grey[900],
    _hover: { bg: dark.grey[200] },
  },
  }
}
```

apps/web/src/styles/theme.ts
```diff
@@ -117,13 +126,59 @@ const theme = extendTheme({
     },
     Button: {
       variants: {
-        solid: {
-          bg: "var(--chakra-colors-gray-100)",
+        solidPrimary: {
```
serjonya-trili, line 129: I think it's better to stick to the names we have in figma: primary, secondary, tertiary

apps/web/src/styles/theme.ts
```diff
@@ -117,13 +126,59 @@ const theme = extendTheme({
...
+        outlinePrimary: {
+          border: "2px solid",
+          borderColor: light.blue,
+          color: light.blue,
+          _hover: {
+            borderColor: light.blueDark,
+            color: light.blueDark,
+          },
+          _dark: {
+            color: "#FFF",
```
serjonya-trili, line 152: dark.grey.black? and in other places too

serjonya-trili (changes requested):

apps/web/src/components/Header/Actions.tsx
```diff
@@ -1,16 +1,18 @@
 import { Flex, IconButton } from "@chakra-ui/react";
+import {} from "@chakra-ui/icons";
```
serjonya-trili, line 2: ?

apps/web/src/styles/useColor.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useCallback } from "react";
+
+import { dark, light } from "./colors";
+
+export const useColor = () => {
+  const colorMode = useColorMode();
+
+  return useCallback(
+    (_light: keyof typeof light.grey, _dark?: keyof typeof dark.grey) =>
+      mode(light.grey[_light], dark.grey[_dark || _light])(colorMode),
```
serjonya-trili, line 12: why only grey? we can apply the same logic to all of the colors

apps/web/src/assets/icons/index.ts
OKendigelyan: They are not in use yet, but ready to be

apps/web/src/styles/useColor.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  useColorMode: jest.fn(),
+}));
+
+jest.mock("../assets/icons", () => ({
+  LogoDarkIcon: () => "logo-dark-icon",
+  LogoLightIcon: () => "logo-light-icon",
+}));
+
+describe("useColor", () => {
+  const mockUseColorMode = useColorMode as jest.Mock;
```
serjonya-trili, line 21: it's better to use jest.mocked(useColorMode) as it is type-safe

apps/web/src/styles/useColor.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const wrapper = ({ children }: PropsWithChildren) => <UmamiTheme>{children}</UmamiTheme>;
```
serjonya-trili, line 23: if you mock useColorMode, do you need to have this wrapper?

apps/web/src/styles/useColor.ts
```diff
@@ -0,0 +1,31 @@
+import { useColorMode } from "@chakra-ui/react";
+import { mode } from "@chakra-ui/theme-tools";
+import { type ReactNode, useCallback } from "react";
+
+import { dark, light } from "./colors";
+
+type ColorScheme = typeof light | typeof dark;
+type GreyScale = 50 | 100 | 200 | 300 | 400 | 450 | 500 | 600 | 700 | 800 | 900 | "black" | "white";
+type Color = keyof Omit<ColorScheme, "grey">;
+type ColorKey = Color | GreyScale | ReactNode;
```
serjonya-trili, line 10: why do we need ReactNode here?

apps/web/src/components/AccountCard/AccountSelector.tsx
```diff
@@ -0,0 +1,49 @@
+import { Flex, IconButton, Text } from "@chakra-ui/react";
+import { useAppSelector } from "@umami/state";
+
+import { ChevronDownIcon, FileCopyIcon, GoogleSocialIcon } from "../../assets/icons";
+import { dark, light } from "../../styles/colors";
+import { useColor } from "../../styles/useColor";
+import { SocialIconWrapper } from "../shared/SocialIconWrapper";
+
+export const AccountSelector = () => {
+  const mode = useColor();
```
serjonya-trili, line 10: I think it is more like const color = useColor(). WDYT?

apps/web/src/styles/useColor.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
serjonya-trili, line 26: IIRC jest does that automatically, no?

OKendigelyan: This only happens if `clearMocks: true` in the global config, and in our case it is true, so we really can remove `jest.clearAllMocks();`

apps/web/src/styles/useColor.test.tsx
OKendigelyan: this is indeed true, but in our case we don't need full type safety, as we will only mock part of the `useColorMode` output

serjonya-trili: okay.png

serjonya-trili (approved):
👍

PR #1651, opened 2024-07-31
Add dynamic disclosure stack

OKendigelyan:
## Proposed changes

This PR implements stack-based disclosure approach allowing to keep state of multiple disclosures such modal windows and drawers and go back to previous disclosure in a stack with the state preserved.

[Task link](https://app.asana.com/0/1207041888417083/1207811126378435/f)

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix

## Checklist


packages/components/src/DynamicDisclosure/DynamicDisclosure.tsx
```diff
@@ -108,51 +77,84 @@ export const useDynamicDisclosure = () => {
...
-    }
+    if (disclosureStack.length === 0) return null;
+
+    return (
+      <>
+        {disclosureStack.map(({ content, props, mode, isOpen }, index) => {
+          const isTopmost = index === disclosureStack.length - 1;
+          const showOverlay =
+            (index > 0 && disclosureStack[index - 1].mode === "drawer") || index === 0;
+          const style = isTopmost ? {} : { display: "none" };
```
serjonya-trili, line 118: I ain't sure if this logic would work tbh.
if we store the stack in a `ref` and keep the current index in the `state` then it should be fine because we don't make any copies of the existing components (e.g. no new component states).
but if we do copy the whole array then the state is going to be replaced with a fresh one once we click the goBack button.

having that given, I think it's not needed to render all of the modals all the time and hiding all but one, we can just render the one that is on that specific index stored in a `useState`. that way, we reuse the same object from memory and it should be fine.

though I might be wrong.

and, in any case, the key must not be `index`. since we don't really know what we store inside, `nanoid` as a key seems to be a good fit

packages/components/src/DynamicDisclosure/DynamicDisclosure.tsx
```diff
@@ -88,82 +52,100 @@ type DynamicDisclosureMode = "modal" | "drawer";
...
+    ...disclosure,
+    content: (
+      <Modal
+        autoFocus={false}
+        blockScrollOnMount={false}
+        closeOnOverlayClick={false}
+        isCentered
+        isOpen={isOpen}
+        motionPreset="slideInBottom"
+        {...props}
```
serjonya-trili, line 113: there should be size="md" as a default parameter

PR #1685, opened 2024-08-05
Add menu items (ui only)

OKendigelyan:
## Proposed changes

[Task link](https://app.asana.com/0/1207041888417083/1207929293130909/f)

This PR just aligns menu UI with figma mockup. Functionality of the separate drawers will be implemented in the separate PR's.

## Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

https://github.com/user-attachments/assets/1dc6da37-0caf-449b-bdcf-834f8416fa22

## Checklist


packages/components/src/DynamicDisclosure/DynamicDisclosure.tsx
```diff
@@ -19,13 +19,18 @@ interface DynamicDisclosureContextType {
   onClose: () => void;
   isOpen: boolean;
   goBack: () => void;
+  stack: Array<{
```
serjonya-trili, line 22: since the only thing you need this for is to check if you need to show a back button, let's have a boolean `hasPreviousPage` or `canGoBack` or whatever. this stack is a too internal thing, and having the same condition implemented in all places we might have a back button is tedious

apps/web/src/components/Onboarding/ModalBackButton.tsx
```diff
@@ -3,7 +3,18 @@ import { IconButton, type IconButtonProps } from "@chakra-ui/react";
 import { ArrowLeftCircleIcon } from "../../assets/icons";
 import { useColor } from "../../styles/useColor";
 
-export const ModalBackButton = (props: IconButtonProps) => {
+export const ModalBackButton = (props: Omit<IconButtonProps, "aria-label">) => {
```
serjonya-trili, line 6: looks like this button will always have `onClick={goBack}` ? maybe it's beneficial to inline it here?

apps/web/src/components/Menu/types.ts
```diff
@@ -0,0 +1,9 @@
+export type TMenuItem = {
+  label: string;
+  icon: React.ReactNode;
```
serjonya-trili, line 3: React is not imported, and it's better to use ReactElement here

apps/web/src/components/Menu/types.ts
```diff
@@ -0,0 +1,9 @@
...
+  onClick: (value?: any) => Promise<void> | void;
```
serjonya-trili, line 4: `() => void` is enough

apps/web/src/components/Menu/types.ts
```diff
@@ -0,0 +1,9 @@
...
+  rightElement?: React.ReactNode;
+  hasArrow?: boolean;
```
serjonya-trili, line 6: maybe it's better to pass in `nextPage` or `link` or whatever here instead? so that the menu item component can decide for itself if it needs an arrow and where this arrow should lead to

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,54 @@
+import { Flex, Icon, Text } from "@chakra-ui/react";
+
+import { ChevronRightIcon } from "../../assets/icons";
+import { useColor } from "../../styles/useColor";
+
+export type TMenuItem = {
```
serjonya-trili, line 6: I definitely saw it [somewhere](https://github.com/trilitech/umami-v2/pull/1685/files#diff-d4717dbfead9ae5b97ceefdb98f6c9288aa208b2a53bed2b891dcfe874f9ae8dR1)

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export type MenuItems = TMenuItem[][];
+
+export const MenuItem = ({ label, icon, onClick, rightElement, hasArrow }: TMenuItem) => {
+  const color = useColor();
+
+  return (
+    <Flex
+      key={label}
```
serjonya-trili, line 21: key normally comes from a parent

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+      alignItems="center"
+      gap="10px"
+      width="full"
+      padding="20px"
+      color="gray.400"
```
serjonya-trili, line 26: please use the useColor here and in other places

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+      _hover={{
+        bg: "gray.100",
+      }}
+      cursor="pointer"
```
serjonya-trili, line 30: so, a flex with onClick and cursor=pointer... isn't it a button? :)

apps/web/src/components/Menu/LogoutModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { useColor } from "../../styles/useColor";
+import { persistor } from "../../utils/persistor";
+import { ModalCloseButton } from "../ModalCloseButton";
+
+export const LogoutModal = () => {
+  const color = useColor();
+  const handleLogout = () => {
+    persistor.pause();
+    localStorage.removeItem("persist:accounts");
```
serjonya-trili, line 19: and "persist:root" too, but I bet there is a ready to go function already that does handle all of that

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export type TMenuItem = {
+  label: string;
+  icon: React.ReactNode;
+  onClick: (value?: any) => Promise<void> | void;
+  rightElement?: React.ReactNode;
+  hasArrow?: boolean;
+};
+
+export type MenuItems = TMenuItem[][];
```
serjonya-trili, line 14: not sure if this is so helpful type tbh, especially [defined twice ](https://github.com/trilitech/umami-v2/pull/1685/files#diff-d4717dbfead9ae5b97ceefdb98f6c9288aa208b2a53bed2b891dcfe874f9ae8dR9)

apps/web/src/components/Menu/AdvancedMenu.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { GenericMenu } from "./GenericMenu";
+import { NetworkMenu } from "./NetworkMenu";
+import { type MenuItems } from "./types";
+import { AlertCircleIcon, LockIcon, RadioIcon } from "../../assets/icons";
+
+export const AdvancedMenu = () => {
+  const { openWith } = useDynamicDrawerContext();
+
+  const menuItems = useMemo<MenuItems>(
```
serjonya-trili, line 12: memo here is not that useful, all these values are cheap to re-render and the drawer itself is quite static anyway

serjonya-trili (changes requested):

apps/web/src/components/Menu/AddressBookMenu.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { EmptyMessage } from "../EmptyMessage";
+import { ModalCloseButton } from "../ModalCloseButton";
+
+export const AddressBookMenu = () => {
+  const color = useColor();
+
+  return (
+    <DrawerContent>
+      <DrawerBackButton />
+      <ModalCloseButton />
```
serjonya-trili, line 14: ?? :)

apps/web/src/components/Menu/AddressBookMenu.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <DrawerBody paddingTop="90px">
+        <Heading color={color("900")} fontSize="24px">
```
serjonya-trili, line 16: it matches size xl. please use it and in other places too

packages/state/src/beacon/WalletClient.ts
```diff
@@ -10,3 +11,12 @@ export const WalletClient =
         iconUrl: "",
         appUrl: "https://umamiwallet.com/",
       });
+
+export const resetWalletConnection = (persistor: Persistor) =>
```
serjonya-trili, line 15: I think a better name would be like logout because it not only reset the wallet connection, but also logs you out

apps/web/src/components/Menu/Menu.tsx
```diff
@@ -1,116 +1,74 @@
...
+      {
+        label: "Apps",
+        icon: <CodeSandboxIcon />,
+        onClick: () => openDrawer(<AppsMenu />),
+        hasArrow: true,
+      },
+    ],
+    [
+      {
+        label: "Dark Mode",
```
serjonya-trili, line 58: this doesn't work for the light mode

apps/web/src/components/Menu/AddressBookMenu.tsx
OKendigelyan: It doesn't match this size, but I added 2xl to match 24px. 
image

serjonya-trili: sorry, I meant 2xl. thanks

serjonya-trili (approved):
🔥

PR #1783, opened 2024-08-23
Fix clear all mnemonic words button

serjonya-trili:


apps/web/src/components/Onboarding/ImportWallet/SeedPhraseTab.tsx
```diff
@@ -0,0 +1,156 @@
...
+                    },
+                    variant: "mnemonic",
+                    placeholder: `word #${index + 1}`,
+                  }}
+                />
+              </GridItem>
+            ))}
+          </Grid>
+          <Button gap="4px" onClick={clearAll} variant="ghost">
+            <Icon as={CloseIcon} />
```
OKendigelyan, line 146: Wouldn't the Button `leftIcon` prop be more appropriate for placing a CloseIcon?

OKendigelyan (approved):

apps/web/src/components/Onboarding/ImportWallet/SeedPhraseTab.tsx
serjonya-trili: sounds reasonable. might need some theme adjustments though because its default margin is 8px, but we need only 4.. let's use it in our next PRs

PR #1813, opened 2024-08-29
Add create password step for mnemonic user

OKendigelyan:
## Proposed changes

[Task link](https://app.asana.com/0/1207041888417083/1207985676541271/f)

## Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

https://github.com/user-attachments/assets/463c5079-21ef-4413-88e7-280bae05b6f1

## Checklist


packages/core/src/Account.ts
```diff
@@ -19,6 +19,7 @@ export type MnemonicAccount = {
   seedFingerPrint: string;
   address: ImplicitAddress;
   pk: string;
+  isVerified: boolean;
```
serjonya-trili, line 22: I think this one is a bit problematic to handle because there can be only 1 unverified account anyway, but this schema allows for multiple.
what if we just add a new key to the accountsSlice itself? this way we can have only 2 states - either verified or not verified among all accounts so to speak.

then the useIsVerified simplifies to just 

```typescript
const useIsVerified = () => useAppSelector(s => s.accounts.isVerified)
```

instead of 
```typescript
const accountIsVerified = accounts.every(acc => acc.type === "mnemonic" && acc.isVerified);
```

also, with this implementation you forgot to update `restoreFromMnemonic`
whilst with the "global" `isVerified` it would've been something like:
```typescript
if (mode === 'new_mnemonic') {
  dispatch(accountActions.setVerified(false);
}
```
in the SetupPassword modal

serjonya-trili (changes requested):

packages/data-polling/src/useGetOperations.tsx
```diff
@@ -99,13 +102,17 @@ export const useGetOperations = (accounts: Account[]) => {
   handleError(error);
 
   useEffect(() => {
-    const interval = setInterval(() => void fetchPreviousPage(), BLOCK_TIME);
+    if (accountIsVerified) {
```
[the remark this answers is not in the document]
OKendigelyan: `useEffect` triggers when `fetchPreviousPage` is initialised. `enabled` prop doesn't affect this behavior

packages/state/src/hooks/setAccountData.ts
```diff
@@ -53,20 +53,23 @@ export const useRestoreFromMnemonic = () => {
...
     label,
     curve,
+    mode = "mnemonic",
   }: {
     mnemonic: string;
     password: string;
     derivationPathTemplate: string;
     label: string;
     curve: Curves;
+    mode?: "mnemonic" | "new-mnemonic";
```
serjonya-trili, line 63: do we need to pass through this additional logic down the callstack? maybe we could simply pass in the `isVerified` key instead with a defaut value being true so that we don't need to update desktop code?

packages/state/src/migrations.test.ts
```diff
@@ -160,4 +160,22 @@ describe("migrations", () => {
       });
     });
   });
+
+  test("8", () => {
+    expect(
+      accountsMigrations[8]({
+        items: [
+          { ...mockMnemonicAccount(0), isVerified: false },
```
serjonya-trili, line 168: please update the mockMnemonicAccount to accept Partial<MnemonicAccount> as the second argument instead

packages/state/src/slices/accounts/accounts.ts
```diff
@@ -131,6 +131,16 @@ export const accountsSlice = createSlice({
     setCurrent: (state, { payload: address }: { payload: RawPkh | undefined }) => {
       state.current = address;
     },
+    setIsVerified: (
+      state,
+      { payload: { pkh, isVerified } }: { payload: { pkh: RawPkh; isVerified: boolean } }
+    ) => {
+      const account = state.items.find(
+        account => account.type === "mnemonic" && account.address.pkh === pkh
```
serjonya-trili, line 139: there is no need to check the type. pkh is unique across all account types

packages/data-polling/src/useGetOperations.tsx
```diff
@@ -47,12 +48,14 @@ type QueryParams =
  *   hasMore - true if there are more operations to fetch
  *   loadMore - function to load more operations (older ones)
  */
-export const useGetOperations = (accounts: Account[]) => {
+export const useGetOperations = (accounts: (Account | ImplicitAccount)[]) => {
```
serjonya-trili, line 51: Account is a superset of ImplicitAccount

packages/data-polling/src/useGetOperations.tsx
```diff
@@ -47,12 +48,14 @@ type QueryParams =
...
   const network = useSelectedNetwork();
   const dispatch = useAppDispatch();
   const refetchTrigger = useRefetchTrigger();
   const handleError = useReactQueryErrorHandler();
 
+  const isVerified = accounts.every(acc => acc.type === "mnemonic" && acc.isVerified);
```
serjonya-trili, line 57: this breaks the behaviour for non-mnemonic account

I think we shouldn't update this piece at all. if you don't need the operations, just don't use the hook on the page

apps/web/src/components/Onboarding/useCheckVerified.tsx
```diff
@@ -0,0 +1,11 @@
+import { useCurrentAccount } from "@umami/state";
+
+export const useCheckVerified = () => {
+  const account = useCurrentAccount();
+
+  if (account && account.type === "mnemonic") {
```
serjonya-trili, line 6:
```typescript
return account?.type !== 'mnemonic'  || account.isVerified
```

apps/web/src/components/Onboarding/useCheckVerified.tsx
```diff
@@ -0,0 +1,11 @@
+import { useCurrentAccount } from "@umami/state";
+
+export const useCheckVerified = () => {
```
serjonya-trili, line 3: `useIsAccountVerified` maybe?

apps/web/src/components/Onboarding/SetupPassword/SetupPassword.tsx
```diff
@@ -66,43 +75,67 @@ export const SetupPassword = ({ mode }: { mode: "mnemonic" | "secret_key" }) =>
...
+              Set a password to unlock your wallet. Make sure to store your password safely.
+            </Text>
+          )}
         </Center>
       </ModalHeader>
-      <ModalBody>
-        <FormProvider {...form}>
-          <form onSubmit={form.handleSubmit(onSubmit)}>
+      <FormProvider {...form}>
+        <form onSubmit={form.handleSubmit(onSubmit)}>
```
serjonya-trili, line 137: why the nesting was changed?

apps/web/src/components/Onboarding/SetupPassword/SetupPassword.tsx
```diff
@@ -36,7 +39,13 @@ type FormFields = {
   curve: Exclude<Curves, "bip25519">;
 };
 
-export const SetupPassword = ({ mode }: { mode: "mnemonic" | "secret_key" }) => {
+type Mode = "mnemonic" | "secret_key" | "new-mnemonic";
```
serjonya-trili, line 42: nit: maybe preserve snake case for consistency?

packages/core/src/testUtils.tsx
```diff
@@ -125,7 +125,11 @@ export const mockImplicitAccount = (
   }
 };
 
-export const mockMnemonicAccount = (index: number, label?: string): MnemonicAccount => ({
+export const mockMnemonicAccount = (
+  index: number,
+  label?: string,
+  isVerified: boolean = true
```
serjonya-trili, line 131: nit: boolean is redundant

not-nit: please make it accept Partial<MnemonicAccount> instead, to avoid things like
```{...mockMnemonicAccount(0), isVerified: false}```

**Review by serjonya-trili — CHANGES_REQUESTED** [review]

File: packages/state/src/hooks/setAccountData.ts (hunk 3 above)
Comment (OKendigelyan) [reply packages/state/src/hooks/setAccountData.ts:63]: this shouldn't affect desktop code since `mode = mnemonic` by default

Comment (serjonya-trili) [reply packages/state/src/hooks/setAccountData.ts:63]: but why this code even need to know about different onboard modes? all it needs is the `isVerified` field

File: packages/state/src/migrations.test.ts (hunk 4 above)
Comment (OKendigelyan) [reply packages/state/src/migrations.test.ts:168]: this change requires updating mockMnemonicAccount across all codebase, so let it be like this for now

File: packages/data-polling/src/useGetOperations.tsx (hunk 7 above)
Comment (OKendigelyan) [reply packages/data-polling/src/useGetOperations.tsx:57]: we can't use useGetOperations conditionally

File: apps/web/src/components/Onboarding/SetupPassword/SetupPassword.tsx [hunk 13]
```diff
@@ -66,43 +75,67 @@ export const SetupPassword = ({ mode }: { mode: "mnemonic" | "secret_key" }) =>
...
-            allFormValues.secretKeyPassword
-          );
-          await restoreFromSecretKey(secretKey, password, label);
-        }
       }
 
       return onClose();
     });
 
+  const isNewMnemonic = mode === "new_mnemonic";
```
serjonya-trili, line 108: you can reuse this variable 2 times in the onSubmit handler too

packages/state/src/migrations.test.ts
```diff
@@ -160,4 +160,20 @@ describe("migrations", () => {
       });
     });
   });
+
+  test("8", () => {
+    expect(
+      accountsMigrations[8]({
+        items: [
+          mockMnemonicAccount(0, { isVerified: false }),
+          mockMnemonicAccount(1, { isVerified: false }),
```
serjonya-trili, line 169: please consider adding other account types to check they aren't affected

serjonya-trili (approved):
a few nits, but other than that 👍

PR #1913, opened 2024-09-18
Improve Beacon payload parsing

serjonya-trili:
now we:
- can parse encoded operations
- show raw payload if user switches the toggle
- fail parsing if it's not a valid UTF-8


OKendigelyan (approved):
LGTM!

PR #1932, opened 2024-09-23
[embed] Fix build for different environments

asiia-trilitech:
## Proposed changes

Previous PR for updating build did not work as expected, fixed it here

Screenshot 2024-09-24 at 18 05 09

[Security review issue](https://github.com/trilitech/umami-security-review/issues/9)

This is preparation for splitting ghostnet & mainnet environments on Vercel:
- Mainnet will have frame-ancestors for allowlisted dApps only. 
- Ghostnet will work for any url, but for ghostnet requests only.

Once everything is set up and tested, I'll add frame-ancestors to `vercel.mainnet.json`

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|        |     |

## Checklist


serjonya-trili (commented):
those files are empty jsons anyway, why do we need to copy-paste them? and can't it be solved with [env variables](https://vercel.com/docs/projects/environment-variables#environments)?

asiia-trilitech:
@serjonya-trili 

> those files are empty jsons anyway, why do we need to copy-paste them? 

> Once everything is set up and tested, I'll add frame-ancestors to vercel.mainnet.json

If I add it now, it will block ghostnet for all users as the setup is not yet completed (new ghostnet deployment is not yet in use & mainnet deployment will require whitelisting on our side after adding frame-ancestors)

> and can't it be solved with [env variables](https://vercel.com/docs/projects/environment-variables#environments)?

Production and preview can be used the same way as mainnet and ghostnet env vars.

However, in this case we will have to support one preview deployment with a fixed ghostnet url that mirrors main branch state at all times. Also testing mainnet with preview deployments won't be possible 

Having separate vercel projects allows to easily deploy main brach code and test both mainnet and ghostnet with preview deployments before the change goes to production.

sergey-kintsel (approved):

PR #1946, opened 2024-09-25
Replace fingerprint calculation with a secure option

serjonya-trili:
now we don't use SHA256 hashes of mnemonics for grouping mnemonic groups


OKendigelyan (approved):
LGTM!

PR #1959, opened 2024-09-27
feat: WalletConnect integration, part 1, session proposal

dianasavvatina:
This is the first part of the WalletConnect integration. It includes the following components:
 - initiating WalletKit by WalletConnect
 - subscribing to basic events
 - handling session proposal on a basic level

Limitations:
 - all requests are rejected
 - no pairing list
 - no way to disconnect
 - no verification of dapp, no check for scam

## Proposed changes

[Task link]()

## Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Steps to reproduce

1. Rejecting
actions: on dApp - connect, copy link. on Wallet: Connect, reject
result: on Wallet - Approve button is inactive. On reject - modal is closed immediately. on dApp: modal is closed

2. Approving
actions: on dApp - connect, copy link. on Wallet: on Wallet: Connect, select Account, Approve
result: on dApp - connected, the list of actions is shown. on Wallet: the modal is closed
actions: on dApp - request transaction
result: on dApp - transaction is immediately rejected with USER rejected error

To be covered in the next PR, #1985:
 - connected dApp should be shown in the list of dApps
 - add a button for disconnecting dApp in the wallet
 - disconnect dApp when it's disconnected on dApp side

## Screenshots

image
image
image

All requests are rejected with 
```
{
  "code": -32000,
  "message": "User rejected methods."
}
```

## Checklist


apps/web/src/components/WalletConnect/ProjectInfoCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Types
+ */
+interface IProps {
+  metadata: SignClientTypes.Metadata;
+  intention?: string;
+}
+
+/**
+ * Components
```
asiia-trilitech, line 15: Please either remove or extend

apps/web/src/components/WalletConnect/WcProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { formatJsonRpcError } from "@walletconnect/jsonrpc-utils";
+import { getSdkError, parseUri } from "@walletconnect/utils";
+import { type PropsWithChildren, useEffect } from "react";
+
+import SessionProposalModal from "./SessionProposalModal";
+
+export const context = {};
+
+export const WalletConnectProvider = ({ children }: PropsWithChildren) => {
+  const currentAccount = useCurrentAccount();
```
asiia-trilitech, line 13: TODO: replace with acc selector

apps/web/src/components/WalletConnect/ProjectInfoCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+ */
+export default function ProjectInfoCard({ metadata, intention }: IProps) {
+  const { icons, name, url } = metadata;
+
+  return (
+    <div style={{ textAlign: "center" }}>
+      <div>
```
asiia-trilitech, line 22: instead of <div> <span> and some other default tags we use chacra components (e.g., <box>) 

please check other components that we already have or ping in the chat for help

apps/web/src/components/WalletConnect/WcProvider.tsx
asiia-trilitech, line 1: nit: maybe wcProvider.tsx -> WalletConnectProvider.tsx ?

apps/web/src/components/WalletConnect/WcProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+        console.error("Error initializing Web3Wallet:", error);
+      }
+    };
+
+    void initializeWallet();
+  });
+
+  async function onSessionProposal(proposal: WalletKitTypes.SessionProposal) {
+    console.log("Session proposal received", proposal);
+    if (!currentAccount) {
```
asiia-trilitech, line 32: Why do we need `currentAccount` and this check? 
There will be account selector in `SessionProposalModal`

Also I assume this check will always be true anyway

apps/web/src/components/WalletConnect/WcProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      await openWith(modal, {});
+    } catch (error) {
+      console.error("Failed to open SessionProposalModal modal:", error);
+    }
+
+    return () => {};
+  }
+
+  async function onSessionRequest(event: WalletKitTypes.SessionRequest) {
```
asiia-trilitech, line 48: We use arrow functions syntax instead of function declarations:
`const onSessionRequest = (event: WalletKitTypes.SessionRequest) => {...}`

Please update here and in other places in order for the code to have the same style across the codebase

apps/web/src/components/WalletConnect/VerifyInfobox.tsx
asiia-trilitech, line 1: Would be good to check existing modals for dApp requests and see what changes we need for the wallet connect implementation (so that designers could start working on that in advance)

This is the latest design for beacon requests that we have: https://www.figma.com/design/l7N0RuVTDuZd2jP77lNqhU/Web-%26-Embed?node-id=6059-103695&node-type=canvas&t=Sc3FdyWv3zI26Sf4-0

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { OwnedImplicitAccountsAutocomplete } from "../AddressAutocomplete";
+
+/**
+ * Finds the overlap between the chains supported by the wallet and the chains requested by the dApp.
+ * We take the first required chain as Umami supports connection for only one chain per request.
+ */
+function getChainToConfirm(required: (string | string[] | undefined)[]) {
+  // take the first required chain which is supported by the wallet
+  for (const chain of required.flat()) {
+    if (chain && ["tezos:mainnet", "tezos:ghostnet"].includes(chain)) {
```
asiia-trilitech, line 30: We should probably use the actual list of networks here by calling `useAvailableNetworks()`

Also would be better to check by urls not by names - (mainnet & ghostnet are default, but for other networks user can select any name they like)

Could be a TODO for the following changes

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+    getValues,
+    formState: { errors, isValid },
+  } = form;
+
+  const required = Object.entries(proposal.params.requiredNamespaces).map(([key, values]) =>
+    key.includes(":") ? key : values.chains
+  );
+
+  if (required.length !== 1) {
+    console.error("Expected only one required chain", required);
```
asiia-trilitech, line 64: If we have this check for only having 1 required network, should we change `getChainToConfirm()` to accept one chain instead of the list?

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+  if (!chain) {
+    console.error("No supported chain found", required);
+  }
+
+  const onApprove = () =>
+    handleAsyncAction(async () => {
+      const account = getAccount(getValues().address);
+
+      try {
+        if (!chain) {
```
asiia-trilitech, line 77: Why the same chain check here again? Would it be possible to only have it in one place?

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
asiia-trilitech: Product question: If there is a list of requested chains that matches more than one Umami network - should we allow user to choose which network they'd like to use for the connection?

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+          throw new Error("No supported chain found");
+        }
+        const namespaces = buildApprovedNamespaces({
+          proposal: proposal.params,
+          supportedNamespaces: {
+            tezos: {
+              chains: [chain],
+              methods: ["tezos_getAccounts", "tezos_sign", "tezos_send"],
```
asiia-trilitech, line 85: Another product question: maybe we can allow user to select which actions they want to allow for the given dApp?

apps/web/src/components/WalletConnect/ProjectInfoCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <Box textAlign="center">
+      <Box>
+        <Avatar marginX="auto" size="lg" src={icons[0]} />
+      </Box>
+      <Box marginTop={4}>
+        <Text data-testid="session-info-card-text">
+          <Text as="span" fontWeight="bold">
+            {name}
+          </Text>{" "}
+          <br />
```
asiia-trilitech, line 27: Please remove `<br /> `

asiia-trilitech (commented):
Thank you for the changes @dianasavvatina 

I've added some more comments

apps/web/src/components/WalletConnect/VerifyInfobox.tsx
```diff
@@ -0,0 +1,19 @@
+import { Box, Card, HStack, Icon, VStack } from "@chakra-ui/react";
+
+import { AlertCircleIcon } from "../../assets/icons";
+
+export default function VerifyInfobox() {
+  return (
+    <Box textAlign="center">
+      <VStack spacing={4}>
```
serjonya-trili, line 8: please use real pixel values instead. here and in all the other places

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      setIsLoadingApprove(false);
+    }).finally(onClose);
+
+  // Handle reject action
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const onReject = () =>
+    handleAsyncAction(async () => {
+      try {
+        setIsLoadingReject(true);
```
OKendigelyan, line 114: `setIsLoadingReject` could be moved to `finally` block after `catch` inside `handleAsyncAction`
the same could be applied to the approve logic

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+        });
+        console.log("approvedNamespaces", namespaces);
+
+        setIsLoadingApprove(true);
+        await walletKit.approveSession({
+          id: proposal.id,
+          namespaces,
+          sessionProperties: {},
+        });
+        onClose();
```
OKendigelyan, line 99: No need to call `onClose()` here since it is called on the line 107 
the same thing in the reject logic

apps/web/src/components/WalletConnect/VerifyInfobox.tsx
dianasavvatina: no access to it yet. requested

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: dApp in WalletConnect sends the names of the chains. How can they be selected by url? let's take it next steps

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: if the user requests more than one required network, then the answer confirming just one network will be rejected by the WalletConnect SDK. So either Umami starts supporting multiple networks in one request (complex task) or dApps must request exactly one network (ok for now?)

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: previously I had to check if chain is known to check if it's available. Now I need to decide what to show in the modal

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: maybe, but that's a new feature which nobody has requested. I suggest to leave it until somebody explicitly wants it

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
asiia-trilitech: Ah, I see 🤔 

@dianasavvatina Could you please add a comment in the code (that all required networks should be supported on umami side for the request to pass)?

IMO just one network should be OK (at least for now), but worth checking with the rest of the team

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
asiia-trilitech: Ah, it's just console.error(), not throwing an error, got it now!

@dianasavvatina 
Not for this PR probably, but it would be good to check what is the expected behaviour on chains mismatch case (when the request would be rejected on the WalletConnect side) - do we want to show SessionProposalModal, just an error toast or nothing at all?

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
asiia-trilitech: Agree, definitely not needed for this PR

But could be good to discuss it with Aswin and maybe create add a backlog task for later (so that it's not get lost)

apps/web/src/components/WalletConnect/VerifyInfobox.tsx
asiia-trilitech: It's more of a note for the future, for now I think we should keep the current design and focus on the integration bit 🙂

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
asiia-trilitech: If it's just `tezos:mainnet` & `tezos:ghostnet` - then there is no need for the check (for now, while we only support tezos). These two networks are always defined in Umami and can't be edited by users. We just need to decide on behaviour in the case of request with multiple chains

If there could be other network options we need to figure out how to check if they're present or not. 
Names for other (custom) networks are coming from the user and might not match WalletConnect names.

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+        walletKit.on("session_request", event => void onSessionRequest(event));
+      } catch (error) {
+        console.error("Error initializing Web3Wallet:", error);
+      }
+    };
+
+    void initializeWallet();
+  });
+
+  const onSessionProposal = async (event: WalletKitTypes.SessionProposal) => {
```
serjonya-trili, line 29: this and the next functions do not provide much value, just inline

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,218 @@
...
+              </>
+            )}
+          </Box>
+          <Divider />
+          <VerifyInfobox />
+        </RequestModalContainer>
+        <ModalFooter>
+          <Button
+            width="100%"
+            isDisabled={false}
```
serjonya-trili, line 198: this is not needed. it's not disabled by default

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,218 @@
...
+            isLoading={isLoadingReject}
+            onClick={onReject}
+            size="lg"
+          >
+            Reject
+          </Button>
+          <Button
+            width="100%"
+            isDisabled={!!error || !isValid}
+            isLoading={isLoadingApprove}
```
serjonya-trili, line 208: why do we have separate loading states for those?

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,218 @@
...
+            isDisabled={false}
+            isLoading={isLoadingReject}
+            onClick={onReject}
+            size="lg"
+          >
+            Reject
+          </Button>
+          <Button
+            width="100%"
+            isDisabled={!!error || !isValid}
```
serjonya-trili, line 207: just `error || !isValid` is sufficient

serjonya-trili (changes requested):

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: `error` is a `string`. 
`!error` - bool, true for undefined or empty error
`!!error` - bool, true for defined error

I get compilation error if change to `error`

asiia-trilitech (approved):
Thanks @dianasavvatina 🌷 

Approved, but please fix this one: 
https://github.com/trilitech/umami-v2/pull/1959#discussion_r1782995835

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: it could be anything, but it must match the config on the wallet.
This is how the proposal looks like:

image

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
dianasavvatina: I've seen the error in toast

apps/web/src/components/Menu/AppsMenu/AppsMenu.tsx
```diff
@@ -2,21 +2,29 @@ import { Button, Divider, Text } from "@chakra-ui/react";
 import { useAddPeer } from "@umami/state";
 
 import { BeaconPeers } from "../../beacon";
+import { useOnWalletConnect } from "../../WalletConnect";
 import { DrawerContentWrapper } from "../DrawerContentWrapper";
 
 export const AppsMenu = () => {
   const addPeer = useAddPeer();
```
OKendigelyan, line 9: I think it would be better to rename this function to something like `useOnBeaconConnect` to make it clearer what each function is responsible for

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+        status: "info",
+      });
+      await refresh();
+    });
+};
+
+export const useOnWalletConnect = () => {
+  const toast = useToast();
+  return async (uri: string) => {
+    try {
```
OKendigelyan, line 109: Why don't we use `handleAsyncAction` here similarly to other functions?

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+  walletKit,
+} from "@umami/state";
+import { type Network } from "@umami/tezos";
+import { formatJsonRpcError } from "@walletconnect/jsonrpc-utils";
+import { getSdkError } from "@walletconnect/utils";
+import { type PropsWithChildren, useEffect } from "react";
+
+import { SessionProposalModal } from "./SessionProposalModal";
+
+export const context = {};
```
OKendigelyan, line 19: Do we need this?

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useEffect(() => {
+    const initializeWallet = async () => {
+      await createWalletKit();
+      walletKit.on("session_proposal", event => void onSessionProposal(event));
+      walletKit.on("session_request", event => void onSessionRequest(event));
+      walletKit.on("session_delete", event => void onSessionDelete(event));
+    };
+
+    void initializeWallet();
+  });
```
OKendigelyan, line 35: Do we need to call it on each render? 
I think we should init wallet only on page mounting, so please set deps array like

` useEffect(() => {
....
  }, []);`
  
 you might want to read [this](https://react.dev/learn/synchronizing-with-effects#step-2-specify-the-effect-dependencies) to get a better understanding of how the effects work in react

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (proposal: WalletKitTypes.SessionProposal) =>
+    handleAsyncActionUnsafe(async () => {
+      // dApp sends in the session proposal the required networks and the optional networks.
+      // The response must contain all the required networks but Umami supports just one per request.
+      // So if the list of required networks is more than one or the required network is not supported, we can only reject the proposal.
+      const requiredNetworks = Object.entries(proposal.params.requiredNamespaces)
+        .map(([key, values]) => (key.includes(":") ? key : values.chains))
+        .flat();
+
+      if (requiredNetworks.length !== 1 || requiredNetworks[0] === undefined) {
```
OKendigelyan, line 54: If you want to check if `there are more than 1 network or networks array is empty` you can simplify to 

`if (requiredNetworks.length > 1 || !requiredNetworks.length) {}`

OKendigelyan (changes requested):

OKendigelyan (approved):
Approve!

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+      }
+      // dApp is waiting so we need to notify it
+      const response = formatJsonRpcError(id, getSdkError("INVALID_METHOD").message);
+      await walletKit.respondSessionRequest({ topic, response });
+    }
+  };
+
+  useEffect(() => {
+    const initializeWallet = async () => {
+      // create the waalet just once
```
serjonya-trili, line 107: typo

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+      await createWalletKit();
+      walletKit.on("session_proposal", event => void onSessionProposal(event));
+      walletKit.on("session_request", event => void onSessionRequest(event));
+      walletKit.on("session_delete", event => void onSessionDelete(event));
+    };
+    void initializeWallet();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
serjonya-trili, line 114: as I said before, please don't disable this check unless you **really** understand what you're doing

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+  }, []);
+
+  return children;
+};
+
+export const useOnWalletConnect = () => {
+  const { handleAsyncAction } = useAsyncActionHandler();
+  return (uri: string) =>
+    handleAsyncAction(async () => {
```
serjonya-trili, line 123: there is no meaning in having async-await right away. it's just `handleAsyncAction(() => walletKit.pair({ uri }))`

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <VStack align="start" spacing="8px">
+              <HStack>
+                <Icon as={CheckmarkIcon} />
+                <Card marginLeft="8px">View your balance and activity</Card>
+              </HStack>
+              <HStack>
+                <Icon as={CheckmarkIcon} />
+                <Card marginLeft="8px">Send approval requests</Card>
+              </HStack>
+              <HStack color="gray.500">
```
serjonya-trili, line 104: please use the `useColor` hook for colors

apps/desktop/src/setupTests.tsx
```diff
@@ -113,3 +113,18 @@ jest.spyOn(console, "error").mockImplementation((...args) => {
   }
   originalError(...args);
 });
+
+jest.mock("@walletconnect/core", () => {
```
serjonya-trili, line 117: nit:
```typescript
jest.mock("@walletconnect/core", () => ({
  Core: jest.fn().mockImplementation((config) => ({
    projectId: config.projectId,
  })),
}));
```

apps/web/src/components/WalletConnect/ProjectInfoCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <Box textAlign="center">
+      <Box>
+        <Avatar marginX="auto" size="lg" src={icons[0]} />
+      </Box>
+      <Box marginTop="16px">
+        <Card data-testid="session-info-card-text">
+          <Text as="span" fontWeight="bold">
+            {name}
+          </Text>{" "}
```
serjonya-trili, line 26: if this blank space is needed then please put it inside of either of the text tags

apps/web/src/components/WalletConnect/SessionProposalModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+      await walletKit.approveSession({
+        id: proposal.id,
+        namespaces,
+        sessionProperties: {},
+      });
+      onClose();
+    });
+
+  // Handle reject action
```
serjonya-trili, line 75: not a useful comment

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,156 @@
...
+        const session = activeSessions[event.topic];
+
+        console.log("Session request, event, session", event, session);
+        console.log("Session request from dApp", session);
+        toast({
+          description: `Session request from dApp ${session.peer.metadata.name}`,
+          status: "info",
+        });
+        throw new Error("Not implemented");
+      } catch (error) {
```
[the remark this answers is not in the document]
dianasavvatina: Yes. It will trigger request rejection. Later when request handling is implemented, the exception will still lead to reject sent to the dApp. This part is already in place.
If we don't send a reject, dApp will be hanging. The basic implementation waits for 5 minutes before giving up and doesn't allow to send another request. So it's easier to send a reject and unblock.

apps/web/src/components/WalletConnect/WalletConnectProvider.tsx
```diff
@@ -0,0 +1,160 @@
...
+    };
+
+    const initializeWallet = async () => {
+      // create the wallet just once
+      if (!isWalletKitCreated.current) {
+        await createWalletKit();
+        isWalletKitCreated.current = true;
+      }
+      // subscribe after the wallet is created
+      removeListeners();
```
serjonya-trili, line 142: why do you need that?

dianasavvatina: I can see that the object walletKit is created twice and the subscription is done twice. Removing listeners guarantees that there will be exactly one subscription

dianasavvatina: fixed handling of listeners. now I save the emitters returned by walletKit.on() and clear listeners on them

serjonya-trili (approved):

PR #2036, opened 2024-10-31
Add smart rollup address support

OKendigelyan:
## Proposed changes

This PR fixes Smart Rollup addresses parsing

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

## Checklist

### DIFF
--- a/apps/desktop/src/components/AddressPill/AddressPill.tsx
+++ b/apps/desktop/src/components/AddressPill/AddressPill.tsx
@@ -35,6 +35,7 @@ export const AddressPill = memo(
       showIcons,
       addressKind,
       addressAlias,
+      address,
       onClick,
       elementRef,
       isMouseHover,
@@ -109,7 +110,7 @@ export const AddressPill = memo(
               </PopoverBody>
             </PopoverContent>
           </Popover>
-          {showIcons && (
+          {showIcons && address.type !== "smart_rollup" && (
             <RightIcon
               marginRight="4px"
               stroke={colors.gray[300]}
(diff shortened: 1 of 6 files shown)


packages/core/src/Operation.ts
```diff
@@ -137,6 +137,8 @@ export const toLambda = (operation: Operation): MichelsonV1Expression[] => {
             operation.recipient.pkh,
             Number(operation.amount)
           );
+        default:
+          throw new Error(`${operation.recipient.type} is not supported yet`);
```
serjonya-trili, line 141: is it though?

apps/desktop/src/components/AddressPill/AddressPill.tsx
```diff
@@ -109,7 +110,7 @@ export const AddressPill = memo(
               </PopoverBody>
             </PopoverContent>
           </Popover>
-          {showIcons && (
+          {showIcons && address.type !== "smart_rollup" && (
```
serjonya-trili, line 113: maybe we could temporarily use the same icon as for the contract?

OKendigelyan: I think it can be confusing for the user

packages/core/src/Operation.ts
OKendigelyan: sending TEZ to rollup isn't supported at the moment

serjonya-trili (approved):

PR #2141, opened 2024-11-21
Add mac release workflow

OKendigelyan:
## Proposed changes

This PR adds mac release workflow

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,20 +6,58 @@ on:
       - "v*.*.*"
 
 jobs:
-  create-draft-release:
+  build:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
         uses: actions/checkout@v4
 
+      - uses: pnpm/action-setup@v3
+        with:
+          version: 9.9.0
+
+      - name: Cache turbo build
+        uses: actions/cache@v4
+        with:
+          path: .turbo
+          key: turbo-${{ github.sha }}
+          restore-keys: |
+            turbo-
+
+      - name: Setup Node.js environment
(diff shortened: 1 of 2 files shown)


ajinkyaraj-23 (approved):
LGTM

.github/workflows/release.yaml
```diff
@@ -21,36 +34,72 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.18.0
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm build
```
serjonya-trili, line 94: you're building all the projects which is redundant. it's better to run turbo build with `working-directory: apps/desktop`

.github/workflows/release.yaml
```diff
@@ -1,17 +1,30 @@
...
 on:
   push:
-     tags:
-       - 'v*.*.*'
+    tags:
+      - "v*.*.*"
+
 jobs:
-  create-release-linux:
+  create-draft-release:
```
serjonya-trili, line 9: it's better to make the downstream jobs explicitly depend on this one

.github/workflows/release.yaml
ajinkyaraj-23: Yes. Thats right.

ajinkyaraj-23: It does not work. Need to build the dependency projects also. need to run it from root directory.

.github/workflows/release.yaml
ajinkyaraj-23: Its already dependent. see the needs: tag in below jobs

PR #2185, opened 2024-12-03
Add generic error message for send operation

OKendigelyan:
## Proposed changes

This PR replaces detailed error messages with generic ones to hide potential sensitive information and make it more user friendly.


asiia-trilitech (commented):
@OKendigelyan just to doublecheck - do we need to add any changes to the desktop?

apps/web/src/views/Tokens/SendTokensFormPage.tsx
asiia-trilitech, line 1: Why these files are being deleted? Are they not used across the codebase?

https://github.com/trilitech/umami-v2/blob/main/apps/web/src/views/Tokens/Token.tsx#L70

packages/core/src/estimate.ts
```diff
@@ -59,13 +57,14 @@ const estimateToEstimation = (estimate: Estimate): Estimation => ({
...
-export const handleTezError = (err: Error): string => {
+export const handleTezError = (err: Error): string | undefined => {
   if (err.message.includes("subtraction_underflow")) {
     return "Insufficient balance, please make sure you have enough funds.";
   } else if (err.message.includes("contract.non_existing_contract")) {
     return "Contract does not exist, please check if the correct network is selected.";
   } else if (err.message.includes("staking_to_delegate_that_refuses_external_staking")) {
     return "The baker you are trying to stake to does not accept external staking.";
+  } else if (err.message.includes("empty_implicit_delegated_contract")) {
+    return "Emptying an implicit delegated account is not allowed.";
```
asiia-trilitech, line 68: nit: maybe something like 

`Emptying an implicit delegated account is not allowed. End delegation before ?trying? again.`

asiia-trilitech (changes requested):
Thank you @OKendigelyan! 
PR looks good to me - please check the comments & I'm happy to approve after!

apps/web/src/views/Tokens/SendTokensFormPage.tsx
OKendigelyan: they've been absolete

apps/web/src/components/Menu/ErrorLogsMenu/ErrorLogsMenu.tsx
```diff
@@ -58,6 +58,11 @@ export const ErrorLogsMenu = () => {
                     <Text color={color("700")} size="sm">
                       {errorLog.timestamp}
                     </Text>
+                    {errorLog.technicalDetails && (
+                      <Text marginTop="12px" color={color("700")} size="sm">
+                        {errorLog.technicalDetails}
```
ajinkyaraj-23, line 63: isnt technicalDetails same as description(old version) if it is visible in Error logs that would be counterproductive to this security issue.

OKendigelyan: assuming we'll require a password from the user to show errors page it should be quite secure
generic message will be shown in the toast

ajinkyaraj-23: The issue was not of security but exposing the detailed architecture of the app. I believe this should be removed and only description and error code should be shown to the user. What do you think @asiia-trilitech ?

packages/core/src/estimate.ts
ajinkyaraj-23: Changed. Thanks

apps/web/src/components/Menu/ErrorLogsMenu/ErrorLogsMenu.tsx
ajinkyaraj-23: Removed for now. We can expand on handleTezErrors going forwards to cover more types of errors.

ajinkyaraj-23:
Created an issue for "too generic error msgs for input validation/secret key validation in restore secret key form". Can be addressed later.

ajinkyaraj-23 (approved):
LGTM

OKendigelyan:
> [ajinkyaraj-23](/ajinkyaraj-23) merged commit [`b3ef307`](/trilitech/umami-v2/commit/b3ef307246e3ebfeb51d8cc2c1d35489d88b6028) into main [Dec 6, 2024](https://github.com/trilitech/umami-v2/pull/2185#event-15560223264)

fixed [here](https://github.com/trilitech/umami-v2/pull/2210)

PR #2319, opened 2025-01-13
Fix deeplink opening from cold start

OKendigelyan:
This pr fixes a bug where an application won't open if you open an initially closed application via deeplink.


asiia-trilitech (approved):
Thank you for the quick fix, @OKendigelyan 

Tested manually - the issue is fixed

PR #2329, opened 2025-01-20
Make @umami/state package platform agnostic

OKendigelyan:
This PR removes `@chakra-ui` dependencies from `@umami/state` and makes it platform agnostic for web and mobile use.


asiia-trilitech (approved):
Thanks @OKendigelyan - the code looks great 🔥 🔥 🔥

PR #2338, opened 2025-01-24
[Web][UMA-1034] hints in modals for Beacon and WC

dianasavvatina:
## Proposed changes

Umami modals for both Beacon and WalletConnect should be more helpful.
A user can easily make a mistake or hesitate confirming a transaction without understanding of the consequences of the operation.

Task: [UMA-1034](https://linear.app/tezos/issue/UMA-1034)

## Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] UI fix

## Steps to reproduce

1. Connect https://taquito-test-dapp.pages.dev/
2. run operations:
 - delegate
 - undelegate
 - stake
 - unstake
 - finalize
3. see the hint and the details description

Other operations are covered in #2340

## Screenshots

| Operation | Before | Now |
| ------ | ------ | --- |
| DelegationRequest | image  | image |
| accordion opened |        | image |
| Undelegation Request | image | image |
| accordion opened |        | image |
| Stake | image | image |
| accordion opened |        | image |
| Unstake | image | image |
| accordion opened |        | image |
| Finalize Unstake Request | image | image |
| accordion opened |        | image |

## Checklist

### DIFF
--- a/apps/web/src/components/HintsAccordion/HintsAccordion.tsx
+++ b/apps/web/src/components/HintsAccordion/HintsAccordion.tsx
@@ -1,0 +1,45 @@
+import {
+  Accordion,
+  AccordionButton,
+  AccordionIcon,
+  AccordionItem,
+  AccordionPanel,
+  Heading,
+} from "@chakra-ui/react";
+import { Hints, type SignPage } from "@umami/core";
+
+import { useColor } from "../../styles/useColor";
+
+type HintsProps = {
+  signPage: SignPage;
+};
+
+export const HintsAccordion = ({ signPage }: HintsProps) => {
+  const color = useColor();
+
+  if (!(signPage in Hints) || !Hints[signPage].header || !Hints[signPage].description) {
+    return null;
+  }
+
+  return (
+    <Accordion
+      width="full"
+      marginTop="16px"
+      marginBottom="16px"
+      allowToggle
+      data-testid="hints-accordion"
+    >
+      <AccordionItem border="none" borderRadius="8px" backgroundColor={color("100")}>
+        <Heading>
+          <AccordionButton padding="12px" borderRadius="8px">
+            <Heading flex="1" textAlign="left" size="md">
+              {Hints[signPage].header}
+            </Heading>
+            <AccordionIcon />
+          </AccordionButton>
+        </Heading>
+        <AccordionPanel padding="16px">{Hints[signPage].description}</AccordionPanel>
+      </AccordionItem>
+    </Accordion>
+  );
+};
--- a/apps/web/src/components/HintsAccordion/index.tsx
+++ b/apps/web/src/components/HintsAccordion/index.tsx
@@ -1,0 +1,1 @@
+export * from "./HintsAccordion";
--- a/apps/web/src/components/SendFlow/common/BatchSignPage.test.tsx
+++ b/apps/web/src/components/SendFlow/common/BatchSignPage.test.tsx
@@ -3,6 +3,7 @@ import type { BatchWalletOperation } from "@taquito/taquito/dist/types/wallet/ba
 import {
   type EstimatedAccountOperations,
   type Operation,
+  Titles,
   executeOperations,
   mockContractCall,
   mockImplicitAccount,
@@ -23,7 +24,6 @@ import {
 import { SuccessStep } from "../SuccessStep";
 import { type SdkSignPageProps, type SignHeaderProps } from "../utils";
 import { BatchSignPage } from "./BatchSignPage";
-import { Titles } from "../../Titles";
 
 jest.mock("@umami/core", () => ({
   ...jest.requireActual("@umami/core"),
--- a/apps/web/src/components/SendFlow/common/BatchSignPage.tsx
+++ b/apps/web/src/components/SendFlow/common/BatchSignPage.tsx
@@ -12,13 +12,13 @@ import {
   ModalFooter,
   Text,
 } from "@chakra-ui/react";
+import { Titles } from "@umami/core";
 import { FormProvider, useForm } from "react-hook-form";
 
 import { Header } from "./Header";
 import { useColor } from "../../../styles/useColor";
 import { AddressTile } from "../../AddressTile/AddressTile";
 import { JsValueWrap } from "../../JsValueWrap";
-import { Titles } from "../../Titles";
 import { useSignWithBeacon } from "../Beacon/useSignWithBeacon";
 import { SignButton } from "../SignButton";
 import { SignPageFee } from "../SignPageFee";
@@ -42,7 +42,7 @@ export const BatchSignPage = (signProps: SdkSignPageProps) => {
 
   return (
     <FormProvider {...form}>
-      <ModalContent data-testId="BatchSignPage">
+      <ModalContent data-testid="BatchSignPage">
         <form>
           <Header headerProps={signProps.headerProps} title={Titles.BatchSignPage} />
 
--- a/apps/web/src/components/SendFlow/common/ContractCallSignPage.tsx
+++ b/apps/web/src/components/SendFlow/common/ContractCallSignPage.tsx
@@ -11,7 +11,7 @@ import {
   ModalContent,
   ModalFooter,
 } from "@chakra-ui/react";
-import { type ContractCall } from "@umami/core";
+import { type ContractCall, Titles } from "@umami/core";
 import { FormProvider, useForm } from "react-hook-form";
 
 import { Header } from "./Header";
@@ -19,8 +19,8 @@ import { useColor } from "../../../styles/useColor";
 import { AddressTile } from "../../AddressTile/AddressTile";
 import { AdvancedSettingsAccordion } from "../../AdvancedSettingsAccordion";
 import { TezTile } from "../../AssetTiles/TezTile";
+import { HintsAccordion } from "../../HintsAccordion";
 import { JsValueWrap } from "../../JsValueWrap";
-import { Titles } from "../../Titles";
 import { SignButton } from "../SignButton";
 import { SignPageFee } from "../SignPageFee";
 import { type CalculatedSignProps, type SdkSignPageProps } from "../utils";
@@ -47,6 +47,7 @@ export const ContractCallSignPage = ({
       <ModalContent data-testid="ContractCallSignPage">
         <form>
           <Header headerProps={headerProps} title={Titles.ContractCallSignPage} />
+          <HintsAccordion signPage="ContractCallSignPage" />
           <ModalBody>
             <TezTile mutezAmount={mutezAmount} />
 
--- a/apps/web/src/components/SendFlow/common/DelegationSignPage.tsx
+++ b/apps/web/src/components/SendFlow/common/DelegationSignPage.tsx
@@ -1,11 +1,11 @@
 import { Flex, FormLabel, ModalBody, ModalContent, ModalFooter } from "@chakra-ui/react";
-import { type Delegation } from "@umami/core";
+import { type Delegation, Titles } from "@umami/core";
 import { FormProvider, useForm } from "react-hook-form";
 
 import { Header } from "./Header";
 import { AddressTile } from "../../AddressTile/AddressTile";
 import { AdvancedSettingsAccordion } from "../../AdvancedSettingsAccordion";
-import { Titles } from "../../Titles";
+import { HintsAccordion } from "../../HintsAccordion";
 import { SignButton } from "../SignButton";
 import { SignPageFee } from "../SignPageFee";
 import { type CalculatedSignProps, type SdkSignPageProps } from "../utils";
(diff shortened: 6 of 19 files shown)


apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  header?: string;
+  description?: string;
+};
+
+export const Hints: Record<SignPage, HintData> = {
+  TezSignPage: {
+    header:
+      "A confirmation takes 1 block (10 seconds). The transaction becomes final and irreversible after 2 blocks (20 seconds).",
+    description:
+      "A confirmation means that the block is confirmed by the network but the operation can still be canceled in the unlikely case of a crypto attack. Finality means that the operation is irreversible.",
```
germanD, line 23:
```suggestion
      "A confirmation means that the block is confirmed by the network but the operation can still be cancelled due to chain reorganizations. Finality means that the operation is irreversible.",
```
Attack is incorrect here. With Tenderbake, you can have a alternative heads scenario, where a block payload proposal generated on (levl n, round k+1) is not confirmed and the chosen chain is a block on (level n+1, usually round 0) which is  a succesor of (level n,  round k (or earlier)). 

This  might just happen out of slow consensus/ degraded health / network congestion scenarios -- which they don't necessarily entail an attack.

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  },
+  ContractCallSignPage: {
+    header:
+      "A confirmation takes 1 block (10 seconds). The transaction becomes final and irreversible after 2 blocks (20 seconds).",
+    description:
+      "A confirmation means that the block is confirmed by the network but the operation can still be canceled in the unlikely case of a crypto attack. Finality means that the operation is irreversible.",
+  },
```
germanD, line 30: Same remarks above

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  DelegationSignPage: {
+    header: "It takes 4 cycles (~6 days) to start receiving delegation rewards.",
+    description:
+      "Delegation rewards are distributed every 2 cycles (~3 days). Ensure that the delegation fee offered by the baker is less than 100%; otherwise, you will not receive rewards. Delegation means that all your funds are delegated to one baker. The funds are still spendable, and you can cancel or change the delegation at any time.",
```
germanD, line 34:
```suggestion
      "Bakers usually distribute delegation rewards  every cycle (~2.8 days). Ensure that the delegation fee offered by the baker is less than 100%; otherwise, you might not receive rewards. Delegation means that all your funds are delegated to one baker. The funds are still spendable, and you can cancel or change the delegation at any time.",
```
Delegation rewards are allocated to the baker during the cycle, they usually pay them a number of cycles in delay. This is not mandated by the protocol. Bakers might pool rewards and pay them each k cycles, for instance

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  },
+  UndelegationSignPage: {
+    header:
+      "Immediately stops delegation and initiates unstaking. Restoring delegation rewards takes 4 cycles (~6 days). Unstaking takes 4 cycles (~10 days).",
```
germanD, line 38:
```suggestion
      "Immediately stops delegation and initiates unstaking. Restoring delegation rewards takes 4 cycles (~10 days). After 4 cycles (~10 days), unstaked tez can be finalized and made avaialable",
```
I need to confirm the math for the unstakes and delegation.

Also notice that in the case of a change of baker, any current liquid balance will be delegated to the new baker, but unstaked tez will keep being delegated to the old one. This is a bit of a corner case where funds end. delegated to two bakers on an overlap of a few cycles . Which is a headache. And not something you want to add to hint :)

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+    description:
+      "Undelegation is applied immediately, so you will not receive rewards anymore. It takes 4 cycles (~6 days) to start receiving delegation rewards again after a new delegation. By undelegating, you also initiate the unstaking process if you have staked. Unstaking will take approximately 4 cycles (~10 days) to complete, after which your staked funds will be returned to your balance.",
+  },
```
germanD, line 41:
```suggestion
      "Undelegation is applied immediately, and the user will  stop accruing **staking** rewards. However, the user should receive delegation rewards for the current and the next 2 cycles, as the baking rights are computed 2 cycles in advance. By undelegating, you also initiate the unstaking process if you have staked. Unstaking will take approximately 4 cycles (~10 days) to complete, after which your staked funds  can be finalized in order to be made  ",
  },
```

This is my understanding, which is consistent with the assumption that unstaked tez would only count as after they are no longer slashable. Waiting for the answer from L1 team on this one.

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  OriginationOperationSignPage: {
+    header:
+      "A confirmation takes 1 block (10 seconds). The transaction becomes final and irreversible after 2 blocks (20 seconds).",
+    description:
+      "A confirmation means that the block is confirmed by the network but the operation can still be canceled in the unlikely case of a crypto attack. Finality means that the operation is irreversible.",
+  },
```
germanD, line 47: Same remarks above

germanD:
@dianasavvatina  made a first pass with several nits/suggestions. Will make another after these are addressed.

Also, opened https://github.com/trilitech/umami-v2/issues/2339 which would complement this (great) effort to improve UX!

germanD (changes requested):
Made some suggestions.

@dianasavvatina I asked the L1 team to answer some of the doubts regarding unstaked tez. That should help clarify the information for tooltips, and will take that into account for a second round!

This is much welcome UX progress :rocket:

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
ajinkyaraj-23, line 1: This file should be part of packages, as the contents (hints) will be common among web/desktop and other platforms.

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
dianasavvatina: we can crate a web page (maybe it already exists) and link it from the hint. Then we can explain all in details including corner cases

apps/web/src/components/HintsAccordion/HintsAccordion.tsx
```diff
@@ -0,0 +1,96 @@
...
+  UnstakeSignPage: {
+    header: "Unstaking takes 4 cycles (~10 days) to complete.",
+    description:
+      "Unstaking is a process that takes approximately 4 cycles (~10 days) to complete. During this time, your funds are locked, and you will not receive rewards. After the unstaking process is complete, you need to finalize unstaking, after which your funds will be returned to your balance.",
+  },
+  FinalizeUnstakeSignPage: {
+    header: "Finalizing unstake takes 1 block (10 seconds).",
+    description:
+      "Finalizing unstake is a process that takes 1 block (10 seconds) to complete. After finalizing unstake, your funds will be returned to your balance. You can then stake them again or spend them.",
+  },
```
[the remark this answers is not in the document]
dianasavvatina: I fixed the block size. the rest seem to be valid.

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type HintData = {
+  header?: string;
+  description?: string;
+};
+const basicOperationHeader = undefined;
+const basicOperationDescription = undefined;
+
+const finalizationConsequences =
+  "Finalized funds are returned to your balance, allowing you to spend them or stake them again to earn rewards. Delegation rewards for these funds stop once they are finalized.";
```
germanD, line 11: > Delegation rewards for these funds stop once they are finalized.

This is incorrect. Both spendable and unstaked balances are delegated. So moving from one to the other doesn't affect delegations.

```suggestion
  "Finalizing unstaked funds makes them available on your spendable balance. They can now be spent or staken again to earn more rewards.";
```

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+const unstakingConsequences = `The unstaking process takes 4 cycles (~10 days) to unlock your staked funds, making them finalizable on request. You will receive delegation rewards for finalizable funds. ${finalizationConsequences}`;
```
germanD, line 12:
```suggestion
const unstakingConsequences = `Unstaked tez immediately stop accruing staking rewards, but sill give rights to delegation rewards. Unstaked tez remain locked for up to 4 cycles (~11 days), after which they can be finalized on request. ${finalizationConsequences}`;
```
If we wanted to be less contrieved : an unstake request on cycle n because finalizable at the beginning of cycle n+4. This means we need to wait at least 3 full cycles, and up to 4.

See the diagram here

https://docs.tezos.com/using/staking#staking-periods-and-delays

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+  OriginationOperationSignPage: {
+    header: basicOperationHeader,
+    description: basicOperationDescription,
+  },
+  ContractCallSignPage: {
+    header: basicOperationHeader,
+    description: basicOperationDescription,
+  },
+  DelegationSignPage: {
+    header: "It takes 2 cycles (~6 days) to start receiving delegation rewards.",
```
germanD, line 28:
```suggestion
    header: "It takes at least 3 full cycles (~9 days) to start receiving delegation rewards. The actual delay will depend on your chosen baker payout strategy",
```
It takes at least 2 full cycles + we need to wait until the end of the cycles to get rewards:
- user delegates on cycle n.
- delegation is taken into account for cycle n+3, and delegation rewards are allocated to the baker during the cycle
- at the beginning of cycle n+4, baker can pay the rewards for the previous cycle,

Note that many bakers pay delegation rewards with further delays. This is off protocols

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+    description:
+      "Bakers typically distribute delegation rewards every cycle (~3 days). Ensure the delegation fee offered by the baker is less than 100%; otherwise, you may not receive rewards. Delegation means that all your funds are delegated to one baker. The funds remain spendable, and you can cancel or change the delegation at any time.",
+  },
+  UndelegationSignPage: {
+    header: "Stops both delegation and staking. Restoring rewards takes 2 cycles (~6 days).",
+    description: `Undelegation takes effect immediately. However, you will still receive delegation rewards for the current and the next 2 cycles, as baking rights are computed 2 cycles in advance. By undelegating, you also initiate the unstaking process if you have staked. ${unstakingConsequences}`,
+  },
+  StakeSignPage: {
+    header: "You will start receiving rewards within 1 cycle (~3 days).",
```
germanD, line 37:
```suggestion
    header: "You will start receiving rewards immediately (within the current cycle)",
```

You can earn rewards from the same next block.

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+    description: `Your spendable balance will be reduced by the amount staked. You will start accruing staking rewards within 1 cycle (~3 days). Staking rewards are compounded on your staked balance. You can cancel staking any time. ${unstakingConsequences}`,
```
germanD, line 38:
```suggestion
    description: `Your spendable balance will be reduced by the amount staked. You will start accruing staking rewards immediately (within the current cycle). Staking rewards are compounded on your staked balance. You can cancel staking any time. ${unstakingConsequences}`,
```

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,48 @@
...
+  },
+  UnstakeSignPage: {
+    header: "Takes at least 4 cycles (~10 days) to complete.",
```
germanD, line 41:
```suggestion
    header: "Unstake request take effect immediately, but funds remain frozen for up to 4 cycles (~10 days)",
```

germanD (approved):
@dianasavvatina LGTM modulo remarks. Depending on the space, some hints and modals can be expanded. But this is imho enough for the current iteration

packages/core/src/Constants/Hints.ts
OKendigelyan, line 1: lowercase `constants` please

packages/core/src/Constants/Hints.ts
```diff
@@ -0,0 +1,49 @@
+import { type SignPage } from "./Titles";
+
+type HintData = {
+  header?: string;
+  description?: string;
+};
+const basicOperationHeader = undefined;
```
OKendigelyan, line 7: Why not empty string `""`?

OKendigelyan (approved):
LGTM!

PR #2343, opened 2025-01-29
Feat/navigation

mariotrilitech:
## Proposed changes
This is a general refactoring aiming at using Tamagui for the UI styling and extracting components from the screens

## Types of changes
- [X] New feature


.expo/README.md
OKendigelyan, line 1: Please remove this from the root

apps/mobile/app/_layout.tsx
```diff
@@ -23,12 +22,16 @@ export default function RootLayout() {
       <ReactQueryProvider>
         <TamaguiProvider config={tamaguiConfig} defaultTheme={colorScheme!}>
           <Provider store={store}>
-            <PersistGate loading={null} persistor={persistor}>
+            <PersistGate
+              loading={
+                <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
```
OKendigelyan, line 27: Please move this to a separate loading component

apps/mobile/app/authenticated/_layout.tsx
OKendigelyan, line 1: Please rename the `authenticated` directory to `(auth)` so as not to add an extra `/authenticated/` to the navigation path, and we could put everything about the logged in user in the (auth) group

apps/mobile/services/auth/useRedirectToAuthorized.ts
```diff
@@ -10,7 +10,7 @@ export const useRedirectToAuthorized = () => {
 
   useEffect(() => {
     if (currentAccount) {
-      router.replace("/home");
+      router.replace("/authenticated/home");
```
OKendigelyan, line 13: this is not necessary if we're going to use (auth) group routing

apps/mobile/app.json
```diff
@@ -1,6 +1,6 @@
 {
   "expo": {
-    "name": "mobile",
+    "name": "mobile-expo",
```
OKendigelyan, line 3: Please revert back to mobile

apps/mobile/app/authenticated/_layout.tsx
mariotrilitech: this is actually a good point, as we are going to support also web in the future, right?

OKendigelyan: it's not so much web related as it is making it convenient to manage different routers under one group

apps/mobile/app/_layout.tsx
```diff
@@ -6,6 +6,7 @@ import { Provider } from "react-redux";
 import { PersistGate } from "redux-persist/integration/react";
 import { TamaguiProvider } from "tamagui";
 
+import { PersistedLoaderComponent } from "../components/persistLoader/persisteLoader.component";
```
OKendigelyan, line 9: Lets name it as PersistorLoader

apps/mobile/components/accountBadge/AccountBadge.component.tsx
OKendigelyan, line 1: Please use the same naming convention for files and directories in all applications: `AccountBadge` for the directory and `AccountBadge.tsx` for the file to match the component name

apps/mobile/screens/Home/Home.screen.tsx
OKendigelyan, line 1: One more comment about naming convention. Please do not use prefixes like .screen, .component since it's clear by the folder name where they are located

apps/mobile/app/(auth)/index.tsx
```diff
@@ -0,0 +1,5 @@
+import { HomeScreen } from "../../screens/Home/Home.screen";
+
+export default function Index() {
```
OKendigelyan, line 3: Please rename to Home

apps/mobile/app/_layout.tsx
```diff
@@ -23,11 +23,14 @@ export default function RootLayout() {
...
         <TamaguiProvider config={tamaguiConfig} defaultTheme={colorScheme!}>
           <Provider store={store}>
-            <PersistGate loading={null} persistor={persistor}>
+            <PersistGate
+              loading={<PersistedLoaderComponent />}
+              persistor={persistor}
+            >
               <AuthProvider>
                 <Stack screenOptions={{ headerShown: false }}>
-                  <Stack.Screen name="home" />
```
OKendigelyan, line 29: I think you should keep `home` and `login`

apps/mobile/services/auth/useRedirectToAuthorized.ts
```diff
@@ -10,9 +10,9 @@ export const useRedirectToAuthorized = () => {
 
   useEffect(() => {
     if (currentAccount) {
-      router.replace("/home");
```
OKendigelyan, line 13: Why do we need to change these lines?

apps/mobile/screens/Home/Home.screen.tsx
mariotrilitech: That’s a valid point, Oleg. The habit of using .component and .screen comes from experience—often, when searching for a file, the location isn’t always immediately clear, depending on the IDE in use. Having an explicit suffix makes it easier to recognise the type of file at a glance (suppose you need to look up a Home component and you have a Home screen and a Home component, if you look it up you may have two Home.tsx files and you may not know which one is what).

That said, if Tamagui has a strict naming convention that requires avoiding these prefixes, I’m happy to follow it. If there’s official documentation on this, I’d appreciate it if you could point me to it so I can align with the standard.

OKendigelyan: Thanks for your response, Mario! 
My suggestion was mainly about our monorepo naming and code guidelines. I do see how .screen and .component files can be distinguished easily, but I feel like this approach adds a bit of clutter. Plus, most major IDEs, including VS Code and WS, support searching not just by filename but also by dir. So, for example, typing 'Home screen' in the search would specifically bring up Home.tsx from the screens directory, making navigation just as efficient

OKendigelyan (approved):

OKendigelyan (approved):

PR #2378, opened 2025-02-10
[Web][UMA-1066] Add fallback for tz symbol on iOS

OKendigelyan:
## Proposed changes

[Task link 1](https://linear.app/tezos/issue/UMA-1066/account-card-balance-the-tez-symbol-doesnt-showload-correctly-tested)
[Task link 2](https://linear.app/tezos/issue/UMA-1061/activity-tab-on-iosmobile-the-tez-symbol-doesnt-showload-correctly)

Add XTZ as a fallback for ꜩ on IOS.

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] UI fix

### DIFF
--- a/packages/tezos/src/constants.ts
+++ b/packages/tezos/src/constants.ts
@@ -1,4 +1,9 @@
-export const TEZ = "ꜩ";
+const getTezSymbol = () => {
+  const isIOS = /iPad|iPhone|iPod/i.test(navigator.userAgent);
+  return isIOS ? "XTZ" : "ꜩ";
+};
+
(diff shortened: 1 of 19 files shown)


packages/tezos/src/constants.ts
```diff
@@ -1,4 +1,14 @@
-export const TEZ = "ꜩ";
+const getTezSymbol = () => {
+  // This is a workaround to avoid the error "navigator is not defined".
```
asiia-trilitech, line 2: Do we still need this workaround?

PR #2389, opened 2025-02-13
feat: replace desktop e2e action auto trigger on manual trigger

OKendigelyan:
Disables auto trigger on desktop-e2e action


ajinkyaraj-23 (approved):
LGTM

PR #2391, opened 2025-02-13
[Web][UMA-1125] Encrypt localstorage with a session key

ajinkyaraj-23:
Encrypt localstorage with a sample key to check the performance.

### DIFF
--- a/apps/desktop/jest.config.ts
+++ b/apps/desktop/jest.config.ts
@@ -6,4 +6,5 @@ export default {
   testTimeout: 20000,
   bail: false,
   setupFilesAfterEnv: ["<rootDir>/src/setupTests.tsx"],
+  transformIgnorePatterns: ["node_modules/(?!(redux-persist-transform-encrypt)/)"],
 };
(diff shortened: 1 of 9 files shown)


packages/state/src/hooks/getAccountData.ts
```diff
@@ -17,11 +17,11 @@ import { useMultisigAccounts } from "./multisig";
 import { useAppSelector } from "./useAppSelector";
 import { type AccountsState, accountsActions } from "../slices";
 
-export const useSeedPhrases = () => useAppSelector(s => s.accounts.seedPhrases);
+export const useSeedPhrases = () => useAppSelector(s => (s.accounts as AccountsState).seedPhrases);
```
bhr, line 20: Why do we need all the casts? Is the initial type not set up correctly?

ajinkyaraj-23: Fixed by adding @types/redux-persist-encrypt-transform package.

PR #2413, opened 2025-02-24
[Web][UMA-1180] Integrate Sentry to web

OKendigelyan:
## Proposed changes

[Task link](https://linear.app/tezos/issue/UMA-1180/add-sentry-to-umami-web)

Add sentry to the web app

### DIFF
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -14,6 +14,7 @@ import { UmamiTheme } from "./providers/UmamiTheme";
 import { store } from "./utils/store";
 import "./utils/hotjar";
 
+import "./utils/sentry";
 import "./index.scss";
 
 // TODO: Move to a hook in @umami/state
--- a/apps/web/src/utils/sentry.ts
+++ b/apps/web/src/utils/sentry.ts
@@ -1,0 +1,9 @@
+import * as Sentry from "@sentry/react";
+
+import { IS_DEV, SENTRY_DSN_KEY } from "../env";
+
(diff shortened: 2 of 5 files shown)


apps/web/src/main.tsx
```diff
@@ -15,6 +15,7 @@ import { UmamiTheme } from "./providers/UmamiTheme";
 import { persistor } from "./utils/persistor";
 import { store } from "./utils/store";
 
+import "./utils/sentry";
```
ajinkyaraj-23, line 18: Add errors to blacklist in redux-persist, now that sentry is setup

apps/web/src/utils/sentry.ts
```diff
@@ -0,0 +1,9 @@
+import * as Sentry from "@sentry/react";
+
+import { IS_DEV, SENTRY_DSN_KEY } from "../env";
+
+Sentry.init({
```
ajinkyaraj-23, line 5: Did you verify if the errors are getting reported on umami project page. I could not see a single error entry on sentry project.

OKendigelyan: sentry is disabled for dev env

apps/web/src/main.tsx
OKendigelyan: I think we need to keep both local errors and sentry to give user way to see what is going on on their non-custodial

ajinkyaraj-23 (approved):
LGTM

PR #2446, opened 2025-03-05
[Web][UMA-1216] Fix incorrect password error within backup flow

OKendigelyan:
## Proposed changes

[Task link](https://linear.app/tezos/issue/UMA-1216/incorrect-password-error-takes-back-to-the-welcome-screen-when)

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] UI fix

### DIFF
--- a/apps/web/src/views/SessionLogin/SessionLogin.tsx
+++ b/apps/web/src/views/SessionLogin/SessionLogin.tsx
@@ -1,6 +1,11 @@
 import { Button, Center, Flex, type FlexProps, Heading, Icon, Text } from "@chakra-ui/react";
 import { type ImplicitAccount, type SocialAccount } from "@umami/core";
-import { type AccountsState, clearSessionKey, useLoginToWallet } from "@umami/state";
+import {
+  type AccountsState,
+  clearSessionKey,
+  useAsyncActionHandler,
+  useLoginToWallet,
+} from "@umami/state";
 import { useState } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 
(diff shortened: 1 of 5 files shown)


packages/state/src/hooks/session.ts
```diff
@@ -108,3 +109,9 @@ export const useLoginToWallet = (
     handleLogin<T extends ImplicitAccount>(): LoginFn<T>;
   };
 };
+
+export const clearStorage = () => {
+  localStorage.removeItem("persist:accounts");
```
ajinkyaraj-23, line 114: Why not just localStorage.clear(). Any items we want to preserve?

ajinkyaraj-23: Do we have passwordattempts with backup? what happens if we exceed, the backup file is not in our control. You can just do localstorage.clear()

ajinkyaraj-23: Also restoring backup is allowed only at the beginning when everything is already new. So we dont need to preserve anything.

OKendigelyan: > Do we have passwordattempts with backup? what happens if we exceed, the backup file is not in our control. You can just do localstorage.clear()

passwordAttempts tied to the password input, not to the backup

apps/web/src/views/SessionLogin/SessionLogin.tsx
```diff
@@ -38,6 +44,15 @@ export const SessionLogin = () => {
     mode: "onBlur",
   });
 
+  const handleSubmit = () =>
```
ajinkyaraj-23, line 47: Its best to move this refactor into another commit. as it does not relate with backup error.

ajinkyaraj-23 (approved):

PR #2466, opened 2025-03-13
[Web][UMA-1055] Fix NFT filter by ui

OKendigelyan:
## Proposed changes

[Task link](https://linear.app/tezos/issue/UMA-1055/nfts-filter-by-filter-by-and-my-total-nfts-number-on-the-right-hand)

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|image|image|

## Checklist

        Please add descriptions of tests added here. How they verify the feature. 

### DIFF
--- a/apps/web/src/components/Tag/Tag.tsx
+++ b/apps/web/src/components/Tag/Tag.tsx
@@ -1,0 +1,21 @@
+import { Button, type ButtonProps, Icon, Text } from "@chakra-ui/react";
+
+import { CloseIcon } from "../../assets/icons";
+
+type TagProps = {
+  option: [string, string];
+  onClick: () => void;
+} & ButtonProps;
+
+export const Tag = ({ option, onClick, ...props }: TagProps) => (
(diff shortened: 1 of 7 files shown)


alecat88 (commented):
@OKendigelyan this seems more like adding a new feature rather than fixing a bug. 
I think we should really add some tests in order to be sure we don't encounter any issue with these new tags in the future. What do you think?

OKendigelyan:
> @OKendigelyan, this appears to be more of a new feature than a bug fix.
> 
> To ensure we don’t run into issues with these new tags in the future, I believe we should add some tests. What are your thoughts?

It makes sense. Added NFTFilter tests to ensure it works correctly

alecat88 (approved):
Well done! Thanks for adding tests. I believe that if we follow this standard, ensuring every new piece of logic is tested, we can be confident that our code will remain stable in the long run and easier to maintain.

PR #2467, opened 2025-03-13
[Web][UMA-1217] Global menu design change

alecat88:
## Proposed changes

[Task link](https://linear.app/tezos/issue/UMA-1217/implement-the-design-changes-for-the-global-menu)

## Types of changes

- [ ] Bugfix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] UI fix

## Screenshots

Before
127 0 0 1_5174_tokens(Full HD)
Now
127 0 0 1_5173_tokens(Full HD)

Before
127 0 0 1_5174_tokens(Full HD) (3)
Now
127 0 0 1_5173_tokens(Full HD) (3)

Before
127 0 0 1_5174_tokens(Full HD) (2)
Now
127 0 0 1_5173_(Full HD)

Before
127 0 0 1_5174_tokens(Full HD) (1)
Now
127 0 0 1_5173_tokens(Full HD) (1)

## Checklist

       1. Rendering Menu Items in Correct Order
       2. Ensures that clicking 'Lock Umami' clears the session and reloads the window.
       3. Tests the functionality of the **Light Mode** toggle.
       4. Checks the visibility of menu items for unverified and unverified users.
       5. Checks the visibility of menu items for users where one of their account is verified, but the current used is/is-not verified.


apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -116,17 +164,37 @@ describe("<Menu />", () => {
...
+      const sessionStorageClearMock = jest.fn();
+      Object.defineProperty(window, "sessionStorage", {
+        value: {
+          ...window.sessionStorage,
+          clear: sessionStorageClearMock,
+        },
+      });
+      delete (window as any).location;
+      const reloadWindowMock = jest.fn();
+      window.location = { reload: reloadWindowMock } as any;
```
OKendigelyan, line 182: Can't we use `jest.spyOn` on reload or mock location for the whole /web in setupTests.ts?

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -144,22 +212,21 @@ describe("<Menu />", () => {
 
     it("renders menu items correctly", async () => {
```
OKendigelyan, line 213: Looks the name of the test case doesn't match what this test checks for. It rather be names as "Doesn't render unverified menu items"

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -168,90 +235,101 @@ describe("<Menu />", () => {
...
+        getAddressBookButton();
+        getAddAccountButton();
+        getSaveBackupButton();
+        getAppsButton();
+        getLightModeButton();
+        getSignOutButton();
+        getPasswordButton();
+        getErrorLogsButton();
+        getLockUmamiButton();
+        if (isVerifiedSelected) {
```
OKendigelyan, line 270: Please add empty line before if statement

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -168,90 +235,101 @@ describe("<Menu />", () => {
...
+          store.dispatch(accountsActions.setCurrent(account.address.pkh));
+        } else {
+          store.dispatch(accountsActions.setCurrent(unverifiedAccount.address.pkh));
+        }
+      });
+
+      it("renders menu items correctly", async () => {
+        await renderInDrawer(<Menu />, store);
+
+        getAddressBookButton();
```
OKendigelyan, line 261: Although this code does what we expect it to do, it doesn't explicitly check that the element is present on the page, which is a bit confusing. 
I would prefer to use expect(...).toBeVisible() check

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -168,90 +235,101 @@ describe("<Menu />", () => {
...
+        const { openWith } = dynamicDrawerContextMock;
+        jest.spyOn(walletKit, "getActiveSessions").mockImplementation(() => ({}));
+
+        await renderInDrawer(<Menu />, store);
+
+        await user.click(screen.getByText(label));
+        expect(openWith).toHaveBeenCalledWith(<Component />);
+      });
+
+      it("opens Sign Out menu correctly", async () => {
```
OKendigelyan, line 293: I think it can be renamed to "opens LogoutModal modal when Sign Out button is clicked"

apps/web/src/components/Menu/Menu.test.tsx
alecat88: I initially thought that since getBy throws an error if the element is not found, it would be sufficient to confirm that the element is on the screen. However, I agree with you—adding .toBeVisible() will help ensure that the element is not just present but also visible, accounting for potential CSS properties that might hide it. This way, we can guarantee a failing test if the element is not actually visible.

apps/web/src/components/Menu/Menu.test.tsx
alecat88: yes, will do that

apps/web/src/components/Menu/Menu.test.tsx
alecat88: Could you clarify the reasoning behind this choice? Is it a standard practice within the codebase? If so, we should consider adding it to ESLint to ensure consistency and automate enforcement.

apps/web/src/components/Menu/Menu.test.tsx
alecat88: This is an existing test, and it currently performs two checks:
	1.	Verifies that the element for an **unverified** user **exists**.
	2.	Ensures that elements meant only for a **verified** user **do not exist**.

I could split this into two separate tests:
	•	Renders menu items correctly
	•	Doesn’t render menu items for a verified user

However, since this test is fundamentally about ensuring the correct menu items are displayed, splitting it would slightly increase the test runtime. Given that both checks already render the same component, I believe keeping them together is more efficient. Would you be okay with leaving it as is?

apps/web/src/components/Menu/Menu.test.tsx
OKendigelyan: It's just make code easy to read since it visually separates different code parts

apps/web/src/components/Menu/Menu.test.tsx
OKendigelyan: I didn't fully expand the hidden unchanged code to see what this test case refers to.
All is good you can keep as it 👍

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -116,17 +164,37 @@ describe("<Menu />", () => {
...
 
-      await user.click(screen.getByText("Light mode"));
+      await user.click(getLightModeButton());
 
       expect(useColorMode().toggleColorMode).toHaveBeenCalled();
     });
 
+    it("it clears the session and reload the window when 'Lock Umami' is clicked", async () => {
+      const sessionStorageClearMock = jest.fn();
+      Object.defineProperty(window, "sessionStorage", {
```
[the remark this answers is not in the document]
alecat88: Thanks, haven't explored that testSetup file yet, will refactor and add sessionStorage there ;)

apps/web/src/components/Menu/Menu.test.tsx
alecat88: I agree, I'll change the eslint for the entire repo in another PR then, so we can always be sure it is respected automatically

apps/web/src/components/Menu/Menu.test.tsx
alecat88: As for the other comment, I haven’t explored the testSetup file yet, thanks for pointing it out! I’ll refactor and add the window location mock there. 😉

apps/web/src/components/Menu/GenericMenu.tsx
```diff
@@ -13,17 +13,33 @@ export const GenericMenu = ({ title, menuItems }: GenericMenuProps) => (
...
     >
-      {menuItems.map((items, index) => (
-        <Box key={index} width="full">
-          {items.map(item => (
-            <MenuItem key={item.label} {...item} />
-          ))}
-        </Box>
-      ))}
+      {menuItems.map((items, i) => {
+        const isLastMenuGroup = i === menuItems.length - 1;
```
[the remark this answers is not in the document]
alecat88: I wasn’t aware of the _last prop—thanks for pointing it out!

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -47,6 +49,44 @@ jest.mock("@umami/state", () => ({
 let store: UmamiStore;
 const account = mockImplicitAccount(0);
 
+const getAddAccountButton = () => screen.getByRole("button", { name: "Add account" });
```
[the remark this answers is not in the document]
alecat88: Following our discussion in the office, we decided to move these helpers into the test-utils package. Gradually consolidating helpers into this package will help reduce code duplication and improve maintainability.

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -63,23 +103,31 @@ describe("<Menu />", () => {
...
       await renderInDrawer(<Menu />, store);
 
-      expect(screen.getByText("Advanced")).toBeVisible();
-      expect(screen.getByText("Address book")).toBeVisible();
-      expect(screen.getByText("Add account")).toBeVisible();
-      expect(screen.getByText("Save backup")).toBeVisible();
-      expect(screen.getByText("Apps")).toBeVisible();
-      expect(screen.getByText("Light mode")).toBeVisible();
-      expect(screen.getByText("Log out")).toBeVisible();
+      for (let i = 1; i < verifiedMenuItems.length; i++) {
```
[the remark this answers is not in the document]
alecat88: As we discussed on Friday, the try/catch with a thrown error might differ from our usual test approach, but in this case, I believe it helps notify the developer of the correct error. Without explicitly throwing the error, the developer would only see numeric values instead of element names, due to the Node API returning numbers: [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition).

I've moved the logic to the test-utils package, so the test will be easier to read.

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -4,7 +4,14 @@ import { type TMenuItem } from "./types";
...
 import { useColor } from "../../styles/useColor";
 
-export const MenuItem = ({ label, icon, onClick, rightElement, hasArrow }: TMenuItem) => {
+export const MenuItem = ({
+  label,
+  icon,
+  onClick,
+  rightElement,
+  hasArrow,
+  style = {},
```
OKendigelyan, line 13:
```suggestion
...props
```

apps/web/src/components/Menu/MenuItem.tsx
```diff
@@ -15,9 +22,11 @@ export const MenuItem = ({ label, icon, onClick, rightElement, hasArrow }: TMenu
       height="auto"
       padding="20px"
       color={color("400")}
+      _last={{ marginTop: "auto" }}
       data-group
       onClick={onClick}
       variant="dropdownOption"
+      {...style}
```
OKendigelyan, line 29:
```suggestion
      {...props}
```

apps/web/src/assets/icons/shield.svg
```diff
@@ -0,0 +1,3 @@
...
...d="M11.6489 1.06367C11.8753 0.978776 12.1247 0.978776 12.3511 1.06367L20.3511 4.06367C20.7414 4.21003 21 4.58316 21 5V12C21 15.4464 18.7183 18.2003 16.6585 20.0026C15.6076 20.9221 14.5615 21.6408 13.78 22.1292C13.3882 22.3741 13.0603 22.5627 12.8281 22.6913C12.7119 22.7556 12.6194 22.805 12.5547 22.839C12.5223 22.856 12.4968 22.8692 12.4788 22.8784L12.4573 22.8894L12.4508 22.8926L12.4486 22.8937C12.4483 22.8939 12.4472 22.8944 12 22C11.5528 22.8944 11.5525 22.8943 11.5522 22.8941L11.5492 22.8926L11.5427 22.8894L11.5212 22.8784C11.5032 22.8692 11.4777 22.856 11.4453 22.839C11.3806 22.805 11.2881 22.7556 11.1719 22.6913C10.9397 22.5627 10.6118 22.3741 10.22 22.1292C9.43854 21.6408 8.39238 20.9221 7.3415 20.0026C5.28175 18.2003 3 15.4464 3 12V5C3 4.58316 3.25857 4.21003 3.64888 4.06367L11.6489 1.06367ZM12 22L11.5522 22.8941C11.8337 23.0349 12.1657 23.0352 12.4472 22.8944L12 22ZM12 20.8628C12.1916 20.7541 12.4367 20.6103 12.72 20.4333C13.4385 19.9842 14.3924 19.3279 15.3415 18.4974C17.2817 16.7997 19 14.5536 19 12V5.693L12 3.068L5 5.693V12C5 14.5536 6.71825 16.7997 8.6585 18.4974C9.60762 19.3279 10.5615 19.9842 11.28 20.4333C11.5633 20.6103 11.8084 20.7541 12 20.8628Z" fill="#909CAF"/>
```
OKendigelyan, line 2: Please use `currentColor` for fill

apps/web/src/assets/icons/sun.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M12 8C9.79086 8 8 9.79086 8 12C8 14.2091 9.79086 16 12 16C14.2091 16 16 14.2091 16 12C16 9.79086 14.2091 8 12 8ZM6 12C6 8.68629 8.68629 6 12 6C15.3137 6 18 8.68629 18 12C18 15.3137 15.3137 18 12 18C8.68629 18 6 15.3137 6 12Z" fill="#909CAF"/>
```
OKendigelyan, line 2: Please use `currentColor` for fill

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -63,23 +107,25 @@ describe("<Menu />", () => {
   });
 
   describe("when user is verified", () => {
-    it("renders menu items correctly", async () => {
-      await renderInDrawer(<Menu />, store);
+    beforeAll(() => {
+      mockWindowLocation();
+    });
+    afterAll(() => {
+      restoreOriginalWindowLocation();
```
OKendigelyan, line 114: We don't need this, since we are using mocked location everywhere

alecat88: Yes, this is necessary.

I initially tried mocking the window.location inside the setupTest file (which applies to every test), but this caused multiple issues in tests that rely on the original window.location object.

apps/web/src/components/Menu/Menu.test.tsx
```diff
@@ -8,9 +8,29 @@ import {
   useDownloadBackupFile,
   walletKit,
 } from "@umami/state";
+import {
```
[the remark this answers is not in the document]
alecat88: That should be fine, you don’t need to worry about long imports. In VS Code, you can simply **collapse** them, keeping the file clean and focused on the actual logic.

At the end of the day, being pragmatic, you’ll spend most of your time reading and writing the logic below the imports, not the imports themselves. Prioritising readability and maintainability of the main code is more important than obsessing over import length.

packages/test-utils/src/components/buttons.ts
```diff
@@ -0,0 +1,21 @@
+import { screen } from "@testing-library/react";
+
+export const getButtonByName = (name: string) => screen.getByRole("button", { name });
+export const queryButtonByName = (name: string) => screen.queryByRole("button", { name });
+
+export const getAddAccountButton = () => getButtonByName("Add account");
```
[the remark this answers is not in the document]
alecat88: OK, I’ve removed all the getters except for the first two utilities: getButtonByName and queryButtonByName, as we discussed.

apps/web/src/components/Menu/Menu.tsx
```diff
@@ -62,27 +69,58 @@ export const Menu = () => {
...
+  const lock = {
+    label: "Lock Umami",
+    icon: <LockIcon />,
+    onClick: handleLockOut,
+  };
 
-  const coreMenuItems = hasVerified
-    ? [advanced, addressBook, addAccount, backup, apps]
-    : [advanced, addAccount];
+  const coreMenuItems = [
```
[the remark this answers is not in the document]
alecat88: What you’re proposing doesn’t follow the same logic.

You’re using only one boolean, but if you look closely, there are actually two: hadVerified and isSelectedAccountVerified.

Also, keep in mind that this array determines the order in which these elements are rendered.

I think, to do something do what you're proposing, the logic that look closer is the following, but I'm not sure it's easier to read:

```
const coreMenuItems = [
  ...(hasVerified ? [advanced, addressBook, addAccount, backup, apps] : [advanced, addAccount]),
  ...(isSelectedAccountVerified ? [network] : []),
  errorLogs,
];
```

OKendigelyan (approved):
LGTM!
Thanks for fixing this bunch of comments 👍

PR #2470, opened 2025-03-17
[Mobile] Add metro config for redux-persist-transform-encrypt

alecat88:
## Proposed changes

Add some missed metro configuration for the "redux-persist-transform-encrypt" package.

## Types of changes

- [ ] Bugfix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] UI fix

## Screenshots

| Before | Now |
| ------ | --- |
|        |     |

## Checklist

        Please add descriptions of tests added here. How they verify the feature. 


apps/mobile/metro.config.js
```diff
@@ -33,6 +42,7 @@ defaultConfig.resolver = {
   extraNodeModules: {
     "@tzkt/oazapfts": path.resolve(workspaceRoot, "node_modules/@tzkt/oazapfts/lib/"),
     "@chakra-ui/utils": path.resolve(workspaceRoot, "node_modules/@chakra-ui/utils/dist/esm"),
+    // "redux-persist-transform-encrypt": path.resolve(workspaceRoot, "node_modules/redux-persist-transform-encrypt/lib"),
```
OKendigelyan, line 45: Please remove this comment

apps/mobile/.env.example
```diff
@@ -0,0 +1 @@
+EXPO_PUBLIC_WEB3_AUTH_CLIENT_ID=abc123
```
OKendigelyan, line 1: Why do we need this?

OKendigelyan (changes requested):

apps/mobile/metro.config.js
alecat88: sorry it slipped by

apps/mobile/.env.example
alecat88: The .env.example file acts as a template for environment variables, guiding developers on required configurations without exposing sensitive data.

The environment variable EXPO_PUBLIC_WEB3_AUTH_CLIENT_ID is required. Without it, the variable at [this line](https://github.com/trilitech/umami-v2/blob/main/apps/mobile/constants/constants.ts#L17) will be undefined, causing the app to throw an error: `“Invalid constructor params, clientId is required”`
This error occurs when initiating authentication in [this line](https://github.com/trilitech/umami-v2/blob/main/apps/mobile/services/auth/AuthClient.ts#L21).

milad-alizadeh: Forget my ignorance as this is my first day :D. 

From what I can see the https://github.com/torusresearch/CustomAuth is used in this project for the mobile app. The web3AuthClientId is a required configuration key when using @toruslabs/customauth (or other Web3Auth SDKs). It identifies the application to Web3Auth services and allows it to securely interact with their login and key management APIs.

```
const torus = new CustomAuth({
  baseUrl: "http://localhost:3000/serviceworker/",
  network: "testnet", // details for test net
  web3AuthClientId: "YOUR_CLIENT_ID", // get Client ID from Web3Auth Dashboard
});
```

apps/mobile/ios/Podfile.lock
```diff
@@ -233,7 +233,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
```
milad-alizadeh, line 233: I would strongly advise against editing iOS pod files directly. Ideally this file should be gitignored as it gets generated by Expo. Nothing to change in this PR but something to look at going forward

milad-alizadeh (approved):

apps/mobile/ios/Podfile.lock
alecat88: Thanks for the feedback! Just to clarify, the Podfile was not edited manually—Expo automatically updated it due to changes in the Metro config.

Regarding Podfile.lock, while Expo generates dependencies, it’s generally recommended to commit this file to maintain consistency across environments and prevent unexpected dependency mismatches. This ensures all developers and CI/CD pipelines install the same versions.

Apple's official documentation also supports this approach:

"If you use CocoaPods, first make sure you commit both your Podfile and the Podfile.lock file."
🔗 [Apple Documentation: Making Dependencies Available to Xcode Cloud](https://developer.apple.com/documentation/xcode/making-dependencies-available-to-xcode-cloud?utm_source=chatgpt.com)

For additional discussions on this topic:

[Reddit discussion on Podfile.lock](https://www.reddit.com/r/flutterhelp/comments/17ib6ze/to_commit_or_not_that_is_the_question/#:~:text=Podfile.&text=lock%20locks%20in%20the%20exact,building%20against%20the%20same%20codebase.)
[Stack Overflow: Should we commit pod files?](https://stackoverflow.com/questions/45587084/should-we-commit-pod-files-to-version-control-system-git-or-svn)
[Stack Overflow: Why include Podfile.lock in version control?](https://stackoverflow.com/questions/44028199/why-should-you-include-podfile-lock-under-version-control)
Nothing to change in this PR, but definitely something we can review for best practices moving forward! 🚀

apps/mobile/.env.example
OKendigelyan: @milad-alizadeh torus doesn't relate to mobile. For mobile we are using web3auth client

